PR #24, opened 2024-11-14
fix(KFLUXUI-194): update usage of build trigger

CryptoRodeo:
Add a tooltip for build trigger details page.
Remove two columns from the component activity page.

Backport of https://github.com/openshift/hac-dev/pull/993


CryptoRodeo:
After taking a look at the lines from Codecov, I believe that code coverage can be ignored here.

src/components/Components/ComponentsListView/ComponentsListRow.tsx
```diff
@@ -30,14 +27,12 @@ export const getContainerImageLink = (url: string) => {
 
 const ComponentsListRow: React.FC<
   RowFunctionArgs<ComponentWithLatestBuildPipeline, PacStatesForComponents>
-> = ({ obj: component, customData }) => {
+> = ({ obj: component }) => {
```
sahil143, line 30: If we do not need that `customData`, Please remove the data passed to customData in the ComponentListView Table component.

sahil143 (changes requested):

src/components/Components/ComponentsListView/ComponentsListRow.tsx
CryptoRodeo: Great catch! Removed: https://github.com/konflux-ci/konflux-ui/pull/24/files#diff-e0b2228fc4bc131576cdbbe79d50a98ad325b7a7d69d746f3ae5a4a2caf6dfabL255

sahil143 (approved):

PR #25, opened 2024-11-15
feat(KFLUXUI-191): allow users to set context

CryptoRodeo:
Allows users to set context values, both in the edit and create integration page.

By default the 'application' context should be selected when creating a new integration test.

Backport of https://github.com/openshift/hac-dev/pull/997

### DIFF
--- a/src/components/IntegrationTests/ContextSelectList.tsx
+++ b/src/components/IntegrationTests/ContextSelectList.tsx
@@ -1,0 +1,216 @@
+import React, { useState } from 'react';
+import {
+  Select,
+  SelectOption,
+  SelectList,
+  MenuToggle,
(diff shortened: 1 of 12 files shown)


CryptoRodeo:
Code coverage should be safe to ignore, based on what I've seen. Most of the lines it's complaining about were not added in this PR. 

Other lines don't seem to make sense, for example it complains about this line:
```
export const ContextSelectList: React.FC<ContextSelectListProps> = ({
```

Yet that component is imported for it's spec test: https://github.com/konflux-ci/konflux-ui/pull/25/files#diff-bafded342a2a58bdb6604e41999d92487b1389c783dad27db3f76713604c87ffR3

sahil143 (approved):

PR #45, opened 2024-12-03
fix(KFLUXUI-198): component relationship remove button

testcara:
Backports KFLUXBUGS-1197


testcara:
After adding more cases, it seems the project coverage rises but the patch coverage keeps stale.
Currently, the project coverage is higher than the base project coverage. 

When I check the patch coverage report, I feel it does not make sense. 

See the screenshot below:
Screenshot 2024-12-04 at 15 19 19

The marked missing hit lines have been covered during tests and more, I do not think counting empty lines helps.

CryptoRodeo (approved):

PR #55, opened 2024-12-13
fix(KFLUXUI-253): tekton results should filter deleted record out

testcara:
## Fixes 
https://issues.redhat.com/browse/KFLUXUI-253

## Description
When we delete one pipeline task run manually before it is completed,UI should not always keep the obsoleted  run there. This patch would ignore the meaningless unknown tekton results to ensure there is no deleted but 'in process' runs on UI.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Without the patch, the UI would list deleted 'running' pipelineruns.
Screenshot 2024-12-16 at 15 03 19
With the patch, the UI would ignore the deleted 'running' pipelineruns.
Screenshot 2024-12-16 at 15 05 12

## How to test or reproduce?
1. Navigate to 'Activity' of one component. 
2. Rerun some jobs and delete it by oc
3. Refresh the page, the running ones would be not shown

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


src/utils/tekton-results.ts
```diff
@@ -303,6 +303,14 @@ const getFilteredPipelineRuns = (
...
     cacheKey,
   );
+  // When pipelines are running, the etcd would keep their results.
+  // deleting pipelines frome ectd would make unknown tekton results.
+  // Just to get meaningful test runs, we need to filter 'unknown' out.
+  const filteredPipelineRuns = originalPipelineRuns.filter((pipelinerun) => {
+    return pipelinerun.status?.conditions?.every((c) => c.status !== 'Unknown') ?? true;
+  });
+  return [filteredPipelineRuns, list];
+};
```
sahil143, line 313: Is it possible that an unknown status will appear for the pipeline runs that were not deleted? In that case, It's not a good idea to filter out the `Unknown` status pipeline runs.

StanislavJochman: Can you specify why it is not good idea?

testcara: I just improved the filter condition to ensure it is more accurate and added more explanations. Could you help to review again? Thank you!

StanislavJochman (dismissed):

CryptoRodeo (dismissed):

src/utils/tekton-results.ts
```diff
@@ -304,6 +304,29 @@ const getFilteredPipelineRuns = (
...
+  */
+  const filteredPipelineRuns = originalPipelineRuns.filter((pipelinerun) => {
+    return (
+      pipelinerun.status?.conditions?.every(
+        (c) => !(c.status === 'Unknown' && c.type === 'Succeeded' && c.reason === 'Running'),
+      ) ?? true
+    );
+  });
+  return [filteredPipelineRuns, list];
+};
```
sahil143, line 328: It's not a good idea to just filter out the `Unknown` status because When a pipeline run is running, its Condition Succeeded status is initially unknown. If the pipeline is deleted while in this state, it is archived to Tekton results with the same 'unknown' status. So we will end up filtering out the Running pipelines also.

sahil143 (changes requested):

src/utils/tekton-results.ts
testcara: For the regular running pipeline, it would be shown running well.  see the screenshot:
Screenshot 2025-02-04 at 10 57 10
We will not filter out regular running pipelines. You could try to rerun one test run like https://localhost:8080/workspaces/wlin-tenant/applications/test-unstopped-pipeline/pipelineruns/test-28bg4 to check the whole process.

The filter will not affect any regular process.
Just when we delete the running pipeline, the pipeline would be gone  and this is the expected behavior.


BTW, do you have better idea, if you would give more suggestion, let me do more improvement with better solution. 
Is it possible we ask other team to add some item in the general status of the pipeline?

src/utils/tekton-results.ts
```diff
@@ -304,6 +304,29 @@ const getFilteredPipelineRuns = (
...
+      "status": "Running",
+    ...
+  Deleting pipelines from etcd makes the tekton record would be never updated as others.
+  So for those tekton results, it is useless to users and we need to filter them out.
+  Otherwise, these jobs would be always shown as 'Running' and bring unexpected troubles.
+  */
+  const filteredPipelineRuns = originalPipelineRuns.filter((pipelinerun) => {
+    return (
+      pipelinerun.status?.conditions?.every(
+        (c) => !(c.status === 'Unknown' && c.type === 'Succeeded' && c.reason === 'Running'),
```
StanislavJochman, line 323: do not we have enum for this?

src/utils/tekton-results.ts
sahil143: Make sense, only Tekton results are being filtered here. can you add a few test cases for this with running status condition and other status condition.

testcara:
@sahil143 , more status have been added to the test. Thank you for your review. :-)

sahil143 (approved):

PR #60, opened 2025-01-06
fix: unblock component list loading from pipeline runs

sahil143:
https://issues.redhat.com/browse/KONFLUX-5766

The components list page takes too long to load because it waits for all pipeline runs to be fetched. Since an application can have a large number of associated pipeline runs, the UI currently fetches only 30 pipeline runs per request, causing the page to wait until all requests are completed.

Proposed Solution:
The UI should load the page as soon as the components are available and fetch the pipeline runs in the background. Each row should display a loading indicator for the pipeline runs. As soon as the pipeline run for a specific component is fetched, it should be displayed on the UI while the remaining pipeline runs continue to load in the background.

**NOTE**:
Added a script to create large number of components for testing purposes. I will remove the script once the PR is tested and approved.

### DIFF
--- a/src/components/Components/ComponentsListView/ComponentListView.tsx
+++ b/src/components/Components/ComponentsListView/ComponentListView.tsx
@@ -79,23 +79,16 @@ const ComponentListView: React.FC<React.PropsWithChildren<ComponentListViewProps
   const componentPACStates = usePACStatesForComponents(components);
 
   const componentsWithLatestBuild = React.useMemo(() => {
-    if (!componentsLoaded || componentsError || !pipelineRunsLoaded || pipelineRunsError) {
+    if (!componentsLoaded || componentsError) {
       return [];
     }
     return components.map((c) => ({
       ...c,
-      latestBuildPipelineRun: pipelineRuns.find(
+      latestBuildPipelineRun: pipelineRuns?.find(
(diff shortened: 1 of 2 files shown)


JoaoPedroPP (approved):

CryptoRodeo (dismissed):
LGTM, I approve the solution

package.json
```diff
@@ -98,7 +99,8 @@
     "webpack-cli": "^5.1.4",
     "webpack-dev-server": "^5.0.4",
     "webpack-merge": "^6.0.1",
-    "whatwg-fetch": "^3.6.20"
+    "whatwg-fetch": "^3.6.20",
+    "zx": "^8.3.0"
```
CryptoRodeo, line 103: Cool, the [zx package](https://www.npmjs.com/package/zx) looks very useful

testcara:
do we show all pipeline runs of the component to users? with times change, pipeline runs would be more and more but most of historical data are useless to users. How about we just show latest 1 month or 2 months? for others, show link to users directing users to get the raw data the linked api directly?

testcara:
The background fetching is helpful to improve current problem. Good improvement. LGTM.

testcara (approved):
LGTM, I approve it.

PR #62, opened 2025-01-07
fix(KFLUXUI-2): Correct GitHub App to stage environment

JoaoPedroPP:


StanislavJochman:
LGTM

src/hooks/useUIInstance.ts
```diff
@@ -9,16 +9,24 @@ export enum ConsoleDotEnvironments {
...
+
 export const getEnv = (): ConsoleDotEnvironments => ConsoleDotEnvironments.prod;
 
-const internalInstance = (host: string) => (env: 'prod' | 'stage') =>
-  new RegExp(`stone-${env}-([A-Za-z0-9]+).([a-z]+).([a-z0-9]+).openshiftapps.com`, 'g').test(host);
+const internalInstance =
+  (host: string) => (env: EnvironmentShortName.prod | EnvironmentShortName.stage) =>
+    new RegExp(`stone-${env}-([A-Za-z0-9]+).([a-z]+).([a-z0-9]+).openshiftapps.com`, 'g').test(
+      host,
+    );
```
sahil143, line 23: We shouldn't be using the regex to determine the environment. It's not reliable.
UI should read this info from the cluster config map.
example: https://github.com/redhat-appstudio/infra-deployments/pull/5168/files#diff-c0903c42c468382bfdf3dd597dcb646d4fccc9e15eddc7af1ac9ec7d3bd4b065
This is the structure of the config map.

sahil143 (changes requested):

testcara:
LGTM

JoaoPedroPP:
We are not working on this yet

PR #67, opened 2025-01-09
chore(KFLUXUI-256): support component.spec.containerImage as latestImage

testcara:
The integration service recently migrated from using the component's spec.containerImage field to store the latest image to using the component's status.lastPromotedImage field. 

The patch support the component.spec.containerImage as the latestImage when it's valid.

### DIFF
--- a/src/components/ApplicationDetails/__data__/WorkflowComponentsData.ts
+++ b/src/components/ApplicationDetails/__data__/WorkflowComponentsData.ts
@@ -106,7 +106,7 @@ export const mockComponentsData = [
     spec: {
       application: 'test-dev-samples',
       componentName: 'test-dotnet60',
-      containerImage: 'quay.io/redhat-appstudio/user-workload:test-ns-test-dotnet60',
+      containerImage: 'quay.io/redhat-appstudio/user-workload:test-ns-test-dotnet61',
       replicas: 1,
       resources: {
         requests: {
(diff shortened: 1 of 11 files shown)


sahil143 (changes requested):
This change should be backward compatible. Right now I have a component that doesn't have any status. We should check for lastPromotedImage first if that doesn't exist use the spec.containerImage.
image

src/components/ApplicationDetails/__data__/WorkflowComponentsData.ts
```diff
@@ -140,6 +140,7 @@ export const mockComponentsData = [
           type: 'Created',
         },
       ],
+      lastPromotedImage: 'quay.io/redhat-appstudio/user-workload:test-ns-test-dotnet60',
       containerImage: 'quay.io/redhat-appstudio/user-workload:test-ns-test-dotnet60',
```
StanislavJochman, line 144: should not here be 61?

StanislavJochman (approved):

src/components/ApplicationDetails/__data__/WorkflowComponentsData.ts
testcara: It is expected. The status.lastPromotedImage is not always same with spec.containerImage.

src/utils/component-utils.ts
```diff
@@ -45,6 +45,21 @@ export type ComponentBuildStatus = {
...
+ *
+ * The latest container image fieled is likely changed from time to time.
+ * So it is valueable to track it as utils to avoid bringing multiple changes
+ * accross several files for furture possible changes.
+ *
+ */
+export const getLastestImage = (component: ComponentKind) => {
+  const latestImage = component.status?.lastPromotedImage || component.spec?.containerImage;
+  return latestImage;
+};
```
sahil143, line 61:
```suggestion
export const getLastestImage = (component: ComponentKind) => component.status?.lastPromotedImage || component.spec?.containerImage;
```

src/utils/__tests__/component-utils.spec.ts
```diff
@@ -160,4 +161,39 @@ describe('component-utils', () => {
...
   });
+
+  it('should return status.lastPromotedImage when lastPromotedImage is available', () => {
+    const mockComponent = {
+      status: {
+        lastPromotedImage: 'test-url',
+      },
+    } as unknown as ComponentKind;
+
+    expect(renderHook(() => getLastestImage(mockComponent)).result.current).toEqual('test-url');
```
sahil143, line 172: use renderHook only for testing react-hooks, not for functions.

```suggestion
    expect(getLastestImage(mockComponent)).toEqual('test-url');
```

sahil143 (changes requested):

sahil143 (approved):

PR #68, opened 2025-01-09
feat(KFLUXUI-125): allow sorting pipeline runs by status and type

marcin-michal:
## Fixes 
https://issues.redhat.com/browse/KFLUXUI-125

## Description
Before, in the application's activity tab, it was only possible to sort pipeline runs by status, and in the snapshot's pipeline runs by type. Now, it's possible to sort by both of these attributes in either of the page, making it more consistent.

## Type of change

- [X] Feature
- [ ] Bugfix
- [ ] Build related changes

image
image
image

## How to test or reproduce?
Go to the activity tab for any application and view the pipeline runs. To view the snapshots pipeline runs, click on any pipeline run from the previous page, click on the snapshot and select the pipeline runs tab.

## Browser conformance: 

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge


CryptoRodeo:
Overall LGTM. I still need to take a closer look and test the functionality live before final review.
For some reason I cannot connect to the staging environment (browser or Konflux UI proxy).

StanislavJochman (approved):
LGTM

StanislavJochman:
LGTM

src/components/PipelineRun/PipelineRunListView/PipelineRunsListView.tsx
```diff
@@ -107,22 +115,51 @@ const PipelineRunsListView: React.FC<React.PropsWithChildren<PipelineRunsListVie
...
+    [typeFiltersParam],
+  );
+
+  const setTypeFilters = (filters: string[]) => setTypeFiltersParam(filters.join(','));
+
+  const typeFilterObj = React.useMemo(() => {
+    return pipelineRuns.reduce((acc, plr) => {
+      if (customFilter && !customFilter(plr)) {
+        return acc;
+      }
```
sahil143, line 131: This same code is being used in Other component, Create reusable utilities or component for the toolbar

sahil143 (changes requested):
Component Details page and Commits details page also has pipeline runs list page. WE will need to update the same filters their also.

marcin-michal:
> Component Details page and Commits details page also has pipeline runs list page. WE will need to update the same filters their also.

@sahil143 Thanks, did not notice pipeline runs were in the Commit details page as well. By Component Details page, do you mean the `ComponentActivityTab` component or a different one? If so, the filters are already there.

marcin-michal:
@sahil143 please check if decomposition like this is okay, so I can work on unit test

StanislavJochman (approved):

StanislavJochman (approved):
LGTM

src/components/Filter/utils/pipelineruns-filter-utils.ts
```diff
@@ -0,0 +1,56 @@
...
+export const createFilterObj = (
+  items: PipelineRunKind[],
+  keyExtractor: (item: PipelineRunKind) => string | undefined,
+  validKeys: string[],
+  filterFn?: (item: PipelineRunKind) => boolean,
+): { [key: string]: number } => {
+  return items.reduce((acc, item) => {
+    if (filterFn && !filterFn(item)) {
+      return acc;
+    }
```
JoaoPedroPP, line 42: Why do we need this if statement with you are always returning `acc`?

marcin-michal: We are always returing `acc` but the difference is whether the key counter goes up or not. By returning early when filtered, it will stay the same

JoaoPedroPP: But you only modify `acc` when there is `key` inside `validKeys`, if not you return `acc` anyway.

marcin-michal: It is possible that we don't want to count some items, even if `key` is inside `validKeys`. For example, `validKeys` might be `["build", "release", "test"]`, pipeline run is of type `"build"`, but we want to ignore pipeline runs with name `"testing-pipeline"`.

JoaoPedroPP (approved):
LGTM

testcara (approved):
LGTM

PR #79, opened 2025-01-14
fix(KONFLUX-6398): fixes swapped pid and taskRunID

StanislavJochman:


src/shared/components/pipeline-run-logs/logs/logs-utils.ts
```diff
@@ -154,8 +154,8 @@ export const getDownloadAllLogsCallback = (
         allLogs += await getTaskRunLog(
           workspace,
           namespace,
-          pipelineRunUID,
           taskRun?.metadata?.uid,
+          pipelineRunUID,
```
gbenhaim, line 158: since you've changed the order of parameters and arguments as well, doe this PR changes something?

StanislavJochman: Still needs testing because not on all places it is changed. But I do not have access to test it so trying to get that.

sahil143 (approved):

PR #84, opened 2025-01-16
fix: ec regex to extract voilation reports from tkn logs

sahil143:
## Fixes 

https://issues.redhat.com/browse/KONFLUX-5697

## Description

with tekton results upgrade tkn logs format changed due to which logs parsing is failing on UI.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

image

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


CryptoRodeo (approved):

src/components/EnterpriseContract/utils.tsx
```diff
@@ -44,8 +44,26 @@ export const getRuleStatus = (type: ENTERPRISE_CONTRACT_STATUS) => {
...
+
 export const extractEcResultsFromTaskRunLogs = (logs: string): EnterpriseContractResult => {
-  const extractedLogs = logs.match(/(\[report-json\] ).+/g);
-  const json = JSON.parse(extractedLogs.map((l) => l.replace('[report-json] ', '')).join(''));
-  return json;
+  const extractedLogs = logs.match(EC_REPORT_JSON_REGEX);
+  try {
+    return JSON.parse(extractedLogs[0]);
+  } catch {
+    return null;
```
StanislavJochman, line 67: Would not make sense to propagate this higher? And ignore it where it is used?

sahil143: Good catch, We should propagate it and show an error state to user rather than hiding it with an empty state.

src/components/EnterpriseContract/EnterpriseContractTable/EnterpriseContractTable.tsx
```diff
@@ -72,7 +72,7 @@ export const EnterpriseContractTable: React.FC<
     <Table variant="compact">
       <Thead>
         <Tr>
-          <Th width={10} />
+          <Th width={10} aria-label="expand toggle" />
           <Th width={30} sort={getSortParams(1)}>
```
testcara, line 76: all th without aria-label would reports warning in console.

src/components/EnterpriseContract/utils.tsx
testcara: we also have the src/utils/enterprise-contract-utils.tsx  which is same with the one. We just need to keep one.

JoaoPedroPP (approved):
LGTM

UI is working as expected. I was able to the logs and check content in security tab.

PR #99, opened 2025-01-29
fix(KFLUXUI-302): remove the join the waitlist flow from UI

StanislavJochman:
Backport of KFLUXUI-302


src/components/SnapshotDetails/__tests__/EnvironmentProvisionErrorAlert.spec.tsx
```diff
@@ -37,7 +37,8 @@ describe('EnvironmentProvisionErrorAlert', () => {
     expect(screen.getByTestId(/env-provision-err-alert/)).toBeInTheDocument();
     expect(screen.getByText('app-sample-go-basic-enterprise-contract')).toBeInTheDocument();
     expect(screen.getByText(/Snapshot failed to deploy/)).toBeInTheDocument();
-    expect(screen.getByText(/Sep 19, 2023/)).toBeInTheDocument();
+    // date in Sep 19, 2023 format or in 19. 9. 2023 format
```
StanislavJochman, line 40: Fixed test to work on both time formats.

JoaoPedroPP (approved):
LGTM.

Join waitlist button was removed from the UI and its components deleted from the project.

src/components/Overview/IntroBanner.tsx
```diff
@@ -18,7 +18,6 @@ import { AccessReviewResources } from '../../types';
 import { useAccessReviewForModels } from '../../utils/rbac';
 import { ButtonWithAccessTooltip } from '../ButtonWithAccessTooltip';
 import { SignupStatus } from '../SignUp/signup-utils';
-import SignupButton from '../SignUp/SignupButton';
 import { useSignupStatus } from '../SignUp/useSignupStatus';
```
sahil143, line 21: Remove other SignUp related components and utilities.

src/components/SnapshotDetails/__tests__/EnvironmentProvisionErrorAlert.spec.tsx
sahil143: We use Sep 19, 2023 format everywhere. Why do you need 19.9.2023 format?

sahil143 (changes requested):

src/components/SnapshotDetails/__tests__/EnvironmentProvisionErrorAlert.spec.tsx
StanislavJochman: When I run it locally it expects 19.9.2023 format 😄. So I rewrote it to be both.

src/components/Overview/IntroBanner.tsx
StanislavJochman: But we still use this right? Only not through button.

sahil143: No , we dont use it.

src/components/SnapshotDetails/__tests__/EnvironmentProvisionErrorAlert.spec.tsx
ralphbean: It seems like this change cannot possibly be logically related to the change in the PR description.

My two cents: you should separate it into a separate commit and separate PR to be considered independently, otherwise this change will be hung up on wonder and debate over this unexpected extra thing.

(You probably need adjust some of your locale environment variables when running the test to get the original expected date format.)

StanislavJochman: I will do it that way. Reverted to original form.

src/components/Overview/IntroBanner.tsx
```diff
@@ -90,8 +83,8 @@ const IntroBanner: React.FC = () => {
                   </Button>
                 ) : undefined}
               </>
-            )}
-            {signupStatus === SignupStatus.PendingApproval && (
+            }
+            {
```
JoaoPedroPP, line 87: I don't think this block makes sense anymore. It will always display that the team is working to to provide you access.

I would say remove is the best action but maybe better check with design

image

src/components/SnapshotDetails/__tests__/EnvironmentProvisionErrorAlert.spec.tsx
ralphbean: Thanks! I'll resolve this comment, but I'll let others provide final review on the overall change (those who know what they're doing more with this codebase than I do!)

src/components/Overview/IntroBanner.tsx
```diff
@@ -55,7 +48,7 @@ const IntroBanner: React.FC = () => {
             </Text>
           </CardBody>
           <CardBody>
-            {signupStatus === SignupStatus.SignedUp && (
+            {
```
JoaoPedroPP, line 51: For simplicity and style if this piece of code is ever going to be displayed it does not need to be in inside `{}` or empty tags `<></>.`

JoaoPedroPP (commented):
Message to provide support is always on the screen and maybe we can improve style

src/components/Overview/IntroBanner.tsx
sahil143: Remove the inline alert.

JoaoPedroPP (dismissed):
LGTM

The UI is working and displaying the `View applications` only if there are applications.
The codecov/patch check can be ignored since it refers to deleted files

sahil143:
@StanislavJochman Please resolve conflicts

sahil143 (changes requested):
There are unnecessary changes in this PR that are not related and caused by improper git rebase. Please rebase properly

PR #105, opened 2025-02-06
fix: handle null items response

ralphbean:
## Fixes 

None on file.

## Description

If you have access to zero workspaces, then it is possible for the /workspaces query to return `null` for the items instead of `[]`.

This change handles that to return `[]` onto the call.

The app fails later where there is undefined behavior if the user cannot land on a workspace, but will deal with that in a later change.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

None.

## How to test or reproduce?

Try to use the UI pointed at the stage cluster, but with a user with no access to workspaces there.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/k8s/k8s-fetch.ts
+++ b/src/k8s/k8s-fetch.ts
@@ -73,11 +73,13 @@ export const k8sListResource = <TResource extends K8sResourceCommon>({
     true,
   ).then((result) => ({
     ...result,
-    items: result.items.map((i) => ({
-      ...i,
-      apiVersion: result.apiVersion,
-      kind: model.kind,
-    })),
+    items: result.items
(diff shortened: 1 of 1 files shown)


testcara:
LGTM

sahil143 (approved):

sahil143:
> The app fails later where there is undefined behavior if the user cannot land on a workspace, but will deal with that in a later change.

The unintended behavior will be addressed by migrating to namespaces. We will implement a list page that displays an empty state, along with instructions on how to gain access to a namespace.

PR #107, opened 2025-02-07
fix(KFLUXUI-312): add useSearchParamBatch hook for batch params operations

marcin-michal:
## Fixes 
[KFLUXUI-312](https://issues.redhat.com/browse/KFLUXUI-312)

## Description
Hook for batch search params operation, fixes the issue where if multiple params were set separately only one of the operation had effect. Will be used in components in [this PR](https://github.com/konflux-ci/konflux-ui/pull/68)

## Type of change
- [X] Bugfix

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/hooks/useSearchParam.ts
```diff
@@ -45,3 +45,54 @@ export const useSearchParam = (
...
+  );
+
+  const batchUnset = React.useCallback(
+    (names?: string[]) => {
+      const newSearchParams = new URLSearchParams(searchParams);
+      searchParams.forEach((_, name) => {
+        if (!names || names.includes(name)) {
+          newSearchParams.delete(name);
+        }
+      });
```
sahil143, line 91:
```suggestion
      names.forEach((name) => {
          newSearchParams.delete(name);
      });
```

src/hooks/useSearchParam.ts
```diff
@@ -45,3 +45,54 @@ export const useSearchParam = (
...
+          newSearchParams.set(name, newValue);
+        }
+      });
+      setSearchParams(newSearchParams);
+    },
+    [searchParams, setSearchParams],
+  );
+
+  const batchUnset = React.useCallback(
+    (names?: string[]) => {
```
sahil143, line 85:
```suggestion
    (names: string[]) => {
```

src/hooks/useSearchParam.ts
```diff
@@ -45,3 +45,54 @@ export const useSearchParam = (
 
   return [value, set, unset];
 };
+
+export const useSearchParamBatch = (): [
```
sahil143, line 49: Add comment and documentation for hook usage

sahil143, line 49: should we allow passing a set of keys that the hook should manage?

sahil143 (changes requested):

src/hooks/useSearchParam.ts
marcin-michal: Since adding a set of managed params, I think it makes sense to keep the option to unset all the params without specifying them. Only the managed ones will be deleted

sahil143 (changes requested):

sahil143 (approved):

PR #112, opened 2025-02-11
fix(KONFLUX-5956): ensure optional username and mandatory password for secret form

rakshett:
## Fixes 

https://issues.redhat.com/browse/KONFLUX-5956

## Description

 removed required for username and added * for password

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Screenshot 2025-02-11 at 5 55 27 PM
image

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/components/Secrets/utils/secret-validation.ts
```diff
@@ -49,10 +49,6 @@ export const secretFormValidationSchema = yup.object({
     is: SecretTypeDropdownLabel.source,
     then: yup.object({
       authType: yup.string(),
-      username: yup.string().when('authType', {
-        is: SourceSecretType.basic,
-        then: yup.string().required('Required'),
```
sahil143, line 54: only remove the required method from the validation

sahil143 (changes requested):

testcara:
How about change the title to something like 'ensure optional username and mandatory password'? 
and it would be better to add test cases for the changes.

rakshett:
> How about change the title to something like 'ensure optional username and mandatory password'? and it would be better to add test cases for the changes.

@testcara Could you please let me know if we need to make this username optional?
Password is already mandatory which doesn't require any change.

sahil143 (approved):

PR #118, opened 2025-02-18
fix(KFLUXUI-302): remove the join the waitlist flow from UI

StanislavJochman:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/components/Overview/IntroBanner.tsx
```diff
@@ -55,67 +46,38 @@ const IntroBanner: React.FC = () => {
...
+              isDisabled={!canCreate}
+              tooltip="You don't have access to create an application"
+              size="lg"
+              analytics={{
+                link_name: 'create-application',
+              }}
+            >
+              Create application
+            </ButtonWithAccessTooltip>
+            {applicationsLoaded && applications?.length > 0 ? (
```
JoaoPedroPP, line 65: Totally optional style suggestion for readability. You don't need the ternary operator here, you can use all as logical `and` statement `applicationsLoaded && applications?.length > 0 && (...)` and remove the `undefined` fo ternary negative case. The `View my application` will only load when `applicationsLoaded` is true and  `applications?.length` is greater than 0.

JoaoPedroPP (approved):
LGTM

UI is working as expected.

src/components/Overview/IntroBanner.tsx
StanislavJochman: I did not change this. Only deleted part in the middle so git diff considers this as added.

testcara:
LGTM

sahil143 (approved):

PR #121, opened 2025-02-24
chore(KFLUXUI-330): replace deprecated Dropdown with compatible select component

testcara:
## Fixes 
[KFLUXUI-330](https://issues.redhat.com/browse/KFLUXUI-330)

## Description
Components of Patterfly are interdependent. To ensure the compatible, let us replace the deprecated dropdown with the dropdown matched with the current Patternfly version.

The current dropdown removed the usage of DropdownToggle and DropdownButton and support the value for 'onSelect' function. In our BasicDropdown, we define our DropdownToggle component by ourselves.

## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

https://github.com/user-attachments/assets/f192e9ca-a95c-4cec-8597-1607a92f884b

## How to test or reproduce?
Visit the pages and try the dropdown to ensure they work as before.

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ReleaseService/ReleasePlan/TriggerRelease/AddIssueSection/__tests__/StatusDropdown.spec.tsx
+++ b/src/components/ReleaseService/ReleasePlan/TriggerRelease/AddIssueSection/__tests__/StatusDropdown.spec.tsx
@@ -1,13 +1,15 @@
-import { act, fireEvent, screen, waitFor } from '@testing-library/react';
+import { screen, waitFor } from '@testing-library/react';
+import { userEvent } from '@testing-library/user-event';
 import { formikRenderer } from '../../../../../../utils/test-utils';
 import StatusDropdown from '../StatusDropdown';
 
 describe('StatusDropdown', () => {
+  const user = userEvent.setup();
   beforeEach(() => {});
 
   it('should show dropdown options', async () => {
     formikRenderer(<StatusDropdown name="status" />);
-    await act(() => fireEvent.click(screen.getByRole('button')));
+    await user.click(screen.getByText('Select status of isssue'));
     expect(screen.getByRole('menuitem', { name: 'Resolved' })).toBeVisible();
     expect(screen.getByRole('menuitem', { name: 'Unresolved' })).toBeVisible();
   });
@@ -16,17 +18,10 @@ describe('StatusDropdown', () => {
     formikRenderer(<StatusDropdown name="status" />, {
       targets: { application: 'app' },
     });
-    expect(screen.queryByRole('button')).toBeInTheDocument();
-
-    await act(() => fireEvent.click(screen.getByRole('button')));
-
+    await user.click(screen.getByText('Select status of isssue'));
+    await user.click(screen.getByText('Unresolved'));
     await waitFor(() => {
-      expect(screen.getByRole('menu')).toBeInTheDocument();
-      screen.getByText('Unresolved');
-    });
-    await act(() => fireEvent.click(screen.getByText('Unresolved')));
-    await waitFor(() => {
-      expect(screen.getByText('Unresolved'));
+      expect(screen.getByTestId('dropdown-toggle').textContent).toEqual('Unresolved');
     });
   });
 });
--- a/src/shared/components/dropdown/BasicDropdown.scss
+++ b/src/shared/components/dropdown/BasicDropdown.scss
@@ -1,20 +1,5 @@
-.basic-dropdown {
-  min-width: 100%;
-  button {
-    min-width: 100%;
-  }
-
-  &[aria-invalid=true] {
-    --pf-v5-c-dropdown__toggle--before--BorderBottomColor: var(--pf-v5-global--danger-color--100);
-    --pf-v5-c-dropdown__toggle--hover--before--BorderBottomColor: var(--pf-v5-global--danger-color--100);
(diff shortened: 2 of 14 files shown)


testcara:
The scss has been changed a lot. I have not found other methods to keep the new dropdown shown as same as previous. Any suggestions are welcomed. Thank you so much.

src/components/ReleaseService/ReleasePlan/TriggerRelease/AddIssueSection/__tests__/StatusDropdown.spec.tsx
```diff
@@ -16,17 +18,10 @@ describe('StatusDropdown', () => {
...
-
-    await act(() => fireEvent.click(screen.getByRole('button')));
-
-    await waitFor(() => {
-      expect(screen.getByRole('menu')).toBeInTheDocument();
-      screen.getByText('Unresolved');
-    });
-    await act(() => fireEvent.click(screen.getByText('Unresolved')));
+    await user.click(screen.getByText('Select status of isssue'));
+    await user.click(screen.getByText('Unresolved'));
```
StanislavJochman, line 22: why are you getting here by text and down bellow by id?

testcara: User.click is an user event, the direct info for user is text options and user clicks the option to choose it. They do not care about the test id.
And for the Line 24, the option value/text is what we need to check. We do it to ensure the expected option is shown as the toggle value. So we cannot get by text.. here.

StanislavJochman (approved):
LGTM

marcin-michal:
Works as expected, I only have a couple of visual improvements:

1. It would be nice if there was some kind of down arrow or similar, to indicate that it is a drop down and not a text input.
2. Visually, it is not consistent with other input fields (has rounded corners, different border and shadow, etc). Not sure it's worth the time and effort, if it would take too much time to fix though.

src/shared/components/dropdown/BasicDropdown.scss
```diff
@@ -1,20 +1,43 @@
...
+.dropdown-toggle {
+  display: flex;
+  align-items: center;
+  padding: 8px 16px;
+  border: 1px solid #d2d2d2;
+  border-radius: 4px;
+  background-color: white;
+  cursor: pointer;
+  min-height: 36px;
+}
```
sahil143, line 15: We shouldn't be using fixed values for custom styling, Please use PF variables. Also you shouldn't need to override pf, default styling from patternfly should be enough.

sahil143 (changes requested):
Can we use https://v5-archive.patternfly.org/components/menus/select instead of dropdown. It basically the same, uses Menu component internally and is more customizable.

testcara:
> Works as expected, I only have a couple of visual improvements:
> 
>     1. It would be nice if there was some kind of down arrow or similar, to indicate that it is a drop down and not a text input.
> 
>     2. Visually, it is not consistent with other input fields (has rounded corners, different border and shadow, etc). Not sure it's worth the time and effort, if it would take too much time to fix though.

Thank you for the testing. I have updated the patch to ensure it is shown as same before. Would you like to help review again? Thank you so much for the help.

src/shared/components/dropdown/BasicDropdown.scss
testcara: Yeah, you're right. I have replaced formats with the pf style. Could you help to review again?
BTW, it seems I have to add more styles than previous scss.  If there is anything to make it shorter, please help to advice. Thank you so much.

src/shared/components/dropdown/BasicDropdown.scss
```diff
@@ -1,20 +1,69 @@
...
+  line-height: var(--pf-v5-global--LineHeight--sm); // var(--pf-v5-c-dropdown__toggle--LineHeight) is higher.
+  height: var(--pf-v5-global--LineHeight--sm);
+
+  /* Arrow Styles */
+  &::after {
+    content: '▼';
+    font-size: var(--pf-v5-global--FontSize--xs);
+    line-height: var(--pf-v5-global--LineHeight--sm);
+    margin-left: var(--pf-global--spacer--sm);
+    transform: scale(1.2, 0.5);
```
sahil143, line 26: Custom CSS shouldn't be needed for base components. Please use Patternfly components.
this will create problems in future when upgrading to a latest version of patternfly

sahil143 (changes requested):
> Can we use https://v5-archive.patternfly.org/components/menus/select instead of dropdown. It basically the same, uses Menu component internally and is more customizable.

Please explore Select component as a replacement

src/shared/components/dropdown/BasicDropdown.scss
testcara: have migrated to select component. With the select component, we do not need to specific these kinds of format. 
Thank you for pointing out the direction.

marcin-michal (dismissed):
LGTM

rrosatti (dismissed):
LGTM

rakshett:
LGTM

janaki29 (approved):

JoaoPedroPP (approved):
LGTM

UI is working as expected.

StanislavJochman (approved):

PR #122, opened 2025-02-24
feat(KFLUXUI-3): updated title for component creation dialog

rakshett:
## Description

"Create an application" is updated as "Create a Component" when "Add Component" button clicked.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

Screenshot 2025-02-24 at 11 07 44 AM
Screenshot 2025-02-24 at 11 07 59 AM

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/components/ImportForm/ImportForm.tsx
```diff
@@ -1,17 +1,23 @@
...
+  const { isApplication = false } = state || {};
   const applicationName = new URLSearchParams(window.location.search).get('application');
   const applicationBreadcrumbs = useApplicationBreadcrumbs(applicationName);
   return (
     <PageLayout
-      breadcrumbs={[...applicationBreadcrumbs, { path: '#', name: 'Create an application' }]}
-      title="Create an application"
+      breadcrumbs={[
+        ...applicationBreadcrumbs,
+        { path: '#', name: `Create an ${isApplication ? 'Application' : 'Component'}` },
```
testcara, line 18: Both the name and the title enjoy the same content.  so how about set it as a const?

testcara: And more it would be good if cases can be added for this part.

rakshett: Updated

marcin-michal:
When creating a component, the title will say 'Create an Component' which has the wrong article. Add another variable for it so it changes accordingly, otherwise looks good.

rakshett:
> When creating a component, the title will say 'Create an Component' which has the wrong article. Add another variable for it so it changes accordingly, otherwise looks good.

Updated the component

src/components/Applications/ApplicationListView.tsx
```diff
@@ -128,6 +128,7 @@ const ApplicationListView: React.FC<React.PropsWithChildren<unknown>> = () => {
                         <Link
                           {...props}
                           to={IMPORT_PATH.createPath({ workspaceName: namespace })}
+                          state={{ isApplication: true }}
```
sahil143, line 131: when user clicks on create component, we add a query param `application` to the url. Use url param to identify if it's a component flow or application flow.

src/components/ImportForm/ImportForm.tsx
```diff
@@ -1,17 +1,21 @@
...
 import ExternalLink from '../../shared/components/links/ExternalLink';
 import { useApplicationBreadcrumbs } from '../../utils/breadcrumb-utils';
 import PageLayout from '../PageLayout/PageLayout';
 import { GitImportForm } from './GitImportForm';
 
 const ImportForm: React.FC = () => {
+  const { state = {} } = useLocation();
+  const { isApplication = false } = state || {};
+  const applicationTitle = `Create an ${isApplication ? 'Application' : 'Component'}`;
   const applicationName = new URLSearchParams(window.location.search).get('application');
```
sahil143, line 13: use this variable to check for the flow.

sahil143 (changes requested):

testcara (commented):
+1 to Sahil. If we are able to get parameter directly to reach our goal, let us not add new parameters.

src/components/ImportForm/ImportForm.tsx
rakshett: Tried using the variable but it is breaking when we directly select the create an application btn

sahil143: What error are you getting while using that variable?
It should be straightforward
```
const applicationTitle = `Create an ${applicationName ? 'Component' : 'Application'  }`;
```

sahil143 (changes requested):

rakshett:
> +1 to Sahil. If we are able to get parameter directly to reach our goal, let us not add new parameters.

@sahil143 We have create application btn in multiple places. In few places we are not getting the applicationName in the url search  hence, we will require this param

sahil143 (approved):

PR #123, opened 2025-02-24
Add PR check test

Katka92:
Adding PR check scripts and GitHub Actions for image push and PR check. 
Adding login code for locally deployed Konflux(+Konflux-ui), updating tests.

## Fixes 
https://issues.redhat.com/browse/KFLUXUI-163
https://issues.redhat.com/browse/KFLUXUI-162

## Description
Enable PR check in a konflux-ui repo.


e2e-tests/utils/Login.ts
```diff
@@ -9,22 +9,25 @@ export class Login {
...
-    // cy.get(loginPO.username).find('[type="text"]').type(username);
-    // cy.get(loginPO.nextButton).click();
-    // cy.get(loginPO.password).find('[type="password"]').type(password, { log: false });
-    // cy.get(loginPO.loginButton).click();
+    cy.get(stageLoginPO.dex).should('be.visible').click();
+    // disabling as we don't have testing account on stage, manual log in is needed
+    // cy.get(stageLoginPO.loginWithSso).click();
+    // cy.get(stageLoginPO.username).type(username);
+    // cy.get(stageLoginPO.password).type(password, { log: false });
+    // cy.get(stageLoginPO.loginButton).click();
```
JoaoPedroPP, line 17: Are this comments required?

e2e-tests/tests/basic-happy-path.spec.ts
```diff
@@ -1,44 +1,33 @@
...
   const repoOwner = 'redhat-hac-qe';
   const publicRepo = `https://github.com/${repoOwner}/${repoName}`;
-  // const publicRepo = 'https://github.com/redhat-hac-qe/devfile-sample-code-with-quarkus-173218691'
   const componentName: string = Common.generateAppName('java-quarkus');
-  // const componentName = "java-quarkus-173261349"
   const piplinerunlogsTasks = ['init', 'clone-repository', 'build-container', 'show-sbom'];
   const quarkusDeplomentBody = 'Congratulations, you have created a new Quarkus cloud application';
   // this is default option and should be the fastest one
-  const pipeline = 'docker-build-oci-ta';
+  // const pipeline = 'docker-build-oci-ta';
```
JoaoPedroPP, line 29: Do we need this?

.github/workflows/base-test-image.yaml
```diff
@@ -0,0 +1,21 @@
...
+on: workflow_dispatch
+
+jobs:
+  base-image:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v4
+
+      - name: Build and Push Base Image
```
testcara, line 13: Why not fedora?

.github/workflows/base-test-image.yaml
```diff
@@ -0,0 +1,21 @@
...
+        env: 
+          QUAY_ORG: ${{ secrets.QUAY_TEST_ORG }}
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN_TEST }}
+        run: |
+          cd e2e-tests
+          podman login -u="$QUAY_ORG" -p="$QUAY_TOKEN" quay.io
+          podman build -t quay.io/konflux_ui_qe/konflux-ui-tests-base:latest -f BaseContainerfile
```
testcara, line 20: set it as var like TEST_IMAGE?

testcara:
The PR generally looks good to me except some tiny questions:
1. Why we enjoy Ubuntu as the layer image, why not Fedora? Is there any benefits here? 
2. I see we enjoy the env to handle multiple vars used in the workflow. How about we also consider to track image names here or some where like vars.env file. Base image name and e2e image name are used in multiple places. If we could handle them as vars, we are able to avoid updates for further change. However, i guess changing image names is rarely happen, so it is fine to leave it as now.
3. It seems we test 'basic urls', 'create applicant with component and its push', do we have plan to add test and release parts here? Konflux makes it easy to do three things: build, test and release. For E2E, covering the whole workflow is more convincing. i guess the added portion won't significantly affect the timing.  I am not sure whether it makes sense.

e2e-tests/utils/Login.ts
Katka92: Once we have a testing account on stage, we can use this code to log in automatically. Would you prefer me to remove the code for now?

e2e-tests/tests/basic-happy-path.spec.ts
Katka92: The default value for a user (and preferably also for a happy path) is the one commented out. In the current setup of Konflux deployment to Github Action, this pipeline doesn't work. Is it enough to add the more specific comment or should I remove it and create a Jira for it?

.github/workflows/base-test-image.yaml
Katka92: I don't see Fedora in available runners hosted by GitHub.
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

testcara:
@Katka92 it is quite clear and thank you for the answer. No questions from my side.
LGTM.

testcara (approved):

e2e-tests/utils/Login.ts
JoaoPedroPP: Got it, so we can keep them until there

e2e-tests/tests/basic-happy-path.spec.ts
JoaoPedroPP: If we don't use it I would recommend to remove it.

testcara (approved):
LGTM

PR #128, opened 2025-02-27
test(fix): fix check for author to be part of an org

Katka92:
Currently, the PR check fails due to validation that the author of the PR is part of an org. 
Trying to use the same approach as Konflux QE job.

Part of https://issues.redhat.com/browse/KFLUXUI-338


JoaoPedroPP (approved):
LGTM

PR #143, opened 2025-03-05
chore(KFLUXVNGD-222): add tekton-bundle-builder pipeline to konflux ui

avi-biton:
Add `tekton-bundle-builder` pipeline to Konflux UI


sahil143:
Closing this PR because UI fetches the pipelines for import flow from config map now.

PR #135, opened 2025-02-28
feat: display release duration in the release list

ralphbean:
## Fixes 

No issue associated with this that I'm aware of.

## Description

This change adds release duration to the list view of releases.

As a user, I want to quickly see how long my most recent releases are taking in the list so I can compare them with each other. If they're taking longer and longer, I have a problem I need to deal with.

We happen to have a lot of real estate in the release list right now.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## Screen shots / Gifs for design review 

None.

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

PR #153, opened 2025-03-13
Pipelinerun links in release list

ralphbean:
## Fixes 

No issues on file that I'm aware of.

## Description

There are two things changing here:

* First, I added a link to the managed pipeline runs associated with each Release the release list. The most common thing I do in the UI is navigate to the list of releases, click on a release only to find the managed pipelinerun link and click it. I never actually stop on the release details page; maybe 1 in 100 times. By presenting this link on the release list, it will save users a click when they're diving in to troubleshoot something.
* Second, the release service gained a new API some time ago for "tenant release pipelines" but we've just never shown them in the UI. The change here adds a link to the tenant release pipeline in the list (just like I added the managed ones) and it also adds the tenant release pipeline link to the release overview page as well.

Lastly, the tests had some old backwards compatibility support for a now unused `status.processing` field. I dropped that. It is long gone.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

[Screencast From 2025-03-13 17-12-01.webm](https://github.com/user-attachments/assets/1eda8263-717e-4a52-84de-eff9cc28edb7)

## How to test or reproduce?

Create a few Releases in your namespace and see how they look in the releases list.

```yaml
apiVersion: appstudio.redhat.com/v1alpha1
kind: Snapshot
metadata:
  name: snapshot
  namespace: user-ns1
spec:
  application: sample-component
  artifacts: {}
  components:
  - containerImage: quay.io/something/goes/here@sha256:19ad3616216eea07d6f1adb48a774dd61c822a5ae800ef43b65766372ee4869b
    name: sample-component
    source:
      git:
        revision: 58d093417befa6dcfc07a89a65d8f08306cbeb59
        url: https://github.com/something/goes/here
---
apiVersion: appstudio.redhat.com/v1alpha1
kind: Release
metadata:
  name: local-release
  namespace: user-ns1
spec:
  gracePeriodDays: 7
  releasePlan: local-release
  snapshot: snapshot
---
apiVersion: appstudio.redhat.com/v1alpha1
kind: Release
metadata:
  name: release
  namespace: user-ns1
spec:
  gracePeriodDays: 7
  releasePlan: sample-component
  snapshot: snapshot
```

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/Releases/ReleasesListHeader.tsx
+++ b/src/components/Releases/ReleasesListHeader.tsx
@@ -2,12 +2,15 @@ import { SortByDirection, ThProps } from '@patternfly/react-table';
 import { HeaderFunc } from '../../shared/components/table/Table';
 
 export const releasesTableColumnClasses = {
-  name: 'pf-m-width-30 pf-m-width-20-on-xl wrap-column',
-  created: 'pf-m-width-20 pf-m-width-20-on-xl',
-  duration: 'pf-m-hidden pf-m-width-20 pf-m-visible-on-xl',
-  status: 'pf-m-width-20',
-  releasePlan: 'pf-m-width-25',
-  releaseSnapshot: 'pf-m-hidden pf-m-width-25 pf-m-visible-on-xl',
+  name: 'pf-m-width-20  pf-m-width-10-on-xl wrap-column',
+  created: 'pf-m-width-20  pf-m-width-15-on-xl',
+  duration: 'pf-m-width-20  pf-m-width-15-on-xl',
+  status: 'pf-m-width-20  pf-m-width-10-on-xl',
+  releasePlan: 'pf-m-width-20  pf-m-width-10-on-xl',
+  releaseSnapshot: 'pf-m-hidden    pf-m-width-10-on-xl pf-m-visible-on-xl',
+  managedPipelineRun: 'pf-m-hidden    pf-m-width-10-on-xl pf-m-visible-on-xl',
+  tenantPipelineRun: 'pf-m-hidden    pf-m-width-10-on-xl pf-m-visible-on-xl',
+  finalPipelineRun: 'pf-m-hidden    pf-m-width-10-on-xl pf-m-visible-on-xl',
   kebab: 'pf-v5-c-table__action',
 };
 
@@ -59,6 +62,18 @@ const getReleasesListHeader: CreateHeader = (activeIndex, activeDirection, onSor
     {
       title: 'Release Snapshot',
       props: { className: releasesTableColumnClasses.releaseSnapshot },
+    },
+    {
+      title: 'Tenant Pipeline',
+      props: { className: releasesTableColumnClasses.tenantPipelineRun },
+    },
+    {
+      title: 'Managed Pipeline',
+      props: { className: releasesTableColumnClasses.managedPipelineRun },
+    },
+    {
+      title: 'Final Pipeline',
+      props: { className: releasesTableColumnClasses.finalPipelineRun },
     },
     {
       title: ' ',
--- a/src/components/Releases/ReleaseOverviewTab.tsx
+++ b/src/components/Releases/ReleaseOverviewTab.tsx
@@ -19,29 +19,29 @@ import { useReleaseStatus } from '../../hooks/useReleaseStatus';
 import { RouterParams } from '../../routes/utils';
 import { Timestamp } from '../../shared/components/timestamp/Timestamp';
 import { useNamespace } from '../../shared/providers/Namespace';
-import { ReleaseKind } from '../../types/release';
 import { calculateDuration } from '../../utils/pipeline-utils';
+import {
+  getNamespaceAndPRName,
+  getManagedPipelineRunFromRelease,
+  getTenantPipelineRunFromRelease,
+  getFinalPipelineRunFromRelease,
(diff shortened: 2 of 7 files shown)


ralphbean:
This builds on top of and includes commits from https://github.com/konflux-ci/konflux-ui/pull/135.

mmalina:
@ralphbean I like this! Would it make sense to also display the final pipeline? Not sure how common it is and I understand that it would make the rows even wider, but on the other hand having all three would make it more consistent.

ralphbean:
Yes, tbh @mmalina I had forgotten that it exists. Yes, let's add it.

I need to refresh myself on the *order* the pipelines run in. I should be able to find that somewhere here, right? https://konflux-ci.dev/docs/advanced-how-tos/releasing/

I see a section on tenant pipelines but nothing on final pipelines at a first glance.

mmalina:
> Yes, tbh @mmalina I had forgotten that it exists. Yes, let's add it.
> 
> I need to refresh myself on the _order_ the pipelines run in. I should be able to find that somewhere here, right? https://konflux-ci.dev/docs/advanced-how-tos/releasing/

It seems docs might be missing. I asked the team. But it was implemented here: [RELEASE-1053](https://issues.redhat.com/browse/RELEASE-1053) and I found [RELEASE-1363](https://issues.redhat.com/browse/RELEASE-1363) which says:

> Both tenant and final pipeline resources are stored in the same spot as they are the same thing (run in dev workspace with same pipeline params), just one runs before managed pipelinerun and one after

So the order should be: tenant pipeline, managed pipeline, final pipeline

mmalina:
Turns out there is an open PR to add more docs on tenant pipelines: https://github.com/konflux-ci/konflux-ui/pull/153

But no mention of final pipelines - David will add some.

JoaoPedroPP (commented):
I have reproduce the steps and my output is bit different. In my release list I don't see anything in `Managed Pipeline Run` column. 

image

src/utils/pipeline-utils.ts
```diff
@@ -86,7 +86,7 @@ export const getDuration = (seconds: number, long?: boolean): string => {
   return duration.trim();
 };
 
-export const calculateDuration = (startTime: string | number, endTime?: string | number) => {
+export const calculateDuration = (startTime?: string | number, endTime?: string | number) => {
```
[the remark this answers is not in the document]
ralphbean: Oh - that may have been a typo that shouldn't have crept in here. Let's remove it.

ralphbean:
> In my release list I don't see anything in Managed Pipeline Run column.

Oh, I know why! The konflux-ci/konflux-ci installer doesn't create the managed namespaces (where the managed pipelines run) by default. I added those resources first before I started my development.

I just submitted https://github.com/konflux-ci/konflux-ci/pull/1569 to correct that for the future.

JoaoPedroPP (dismissed):
LGTM

I did what you mentioned in konflux-ci repo and I as able to see both types of pipeline.

UI is working as expected

src/components/Releases/ReleasesListHeader.tsx
```diff
@@ -3,10 +3,13 @@ import { HeaderFunc } from '../../shared/components/table/Table';
...
   name: 'pf-m-width-30 pf-m-width-20-on-xl wrap-column',
-  created: 'pf-m-width-30 pf-m-width-20-on-xl',
+  created: 'pf-m-width-20 pf-m-width-20-on-xl',
+  duration: 'pf-m-hidden pf-m-width-20 pf-m-visible-on-xl',
   status: 'pf-m-width-20',
   releasePlan: 'pf-m-width-25',
   releaseSnapshot: 'pf-m-hidden pf-m-width-25 pf-m-visible-on-xl',
+  managedPipelineRun: 'pf-m-hidden pf-m-width-25 pf-m-visible-on-xl',
+  tenantPipelineRun: 'pf-m-hidden pf-m-width-25 pf-m-visible-on-xl',
   kebab: 'pf-v5-c-table__action',
```
sahil143, line 13: Total width of all `pf-m-width-<number>` should be 100, right now it's adding up to 190 which can cause table resizing issues on different viewports.

sahil143 (changes requested):

src/components/Releases/ReleasesListHeader.tsx
ralphbean: Ack thanks - will fix!

ralphbean: Ok, I think I've fixed it.

I put some extra whitespace in to make it easier (for me) to visually confirm that the numbers make sense.

I don't see that style elsewhere in the repo though. Is that kind of whitespace acceptable? If not, I can remove it.

ralphbean: @sahil143 I rebased on `main`.

ralphbean: I was missing an update to `src/types/release.ts`, fixed just now.

sahil143 (approved):
@ralphbean Thanks for the PR. Changes looks good.

ralphbean:
Thanks @sahil143 !

PR #157, opened 2025-03-17
feat(KFLUXUI-343): unify filtering logic

marcin-michal:
[KFLUXUI-343](https://issues.redhat.com/browse/KFLUXUI-343)

use generic filter context and move all filtering logic to there
unify similar filter toolbars to reduce duplicity
make adding new filters easy and uniform


JoaoPedroPP (commented):
So basically what you did was apply the filter you created in previous PR to all components that have a search bar, is that correct?

marcin-michal:
@JoaoPedroPP Basically yes, but I also changed the filter context provider to be more generic, so there is no need to have a specific one for each toolbar type

src/routes/index.tsx
```diff
@@ -52,7 +52,6 @@ export const router = createBrowserRouter([
       ...pipelineRoutes,
       ...taskRunRoutes,
       ...userAccessRoutes,
-
```
testcara, line 55: let us bring it back to avoid unexpected changes.

testcara (dismissed):
UI works as expected. 
It brings some changes like:
Screenshot 2025-05-14 at 16 14 58
The 'all usernames' are not there. 

But for me , it is ok about the change.

testcara (approved):
LGTM

PR #163, opened 2025-03-19
fix(KONFLUX-7308): don't expand/collapse advanced git options while typing url

marcin-michal:
## Fixes 
[KONFLUX-7308](https://issues.redhat.com/browse/KONFLUX-7308)

## Description
When the user inputs the git repo URL, the advanced git options would open/close in a not very intuitive way. Now, the section is closed by default and can be opened only by the user clicking the open button. 

## Type of change

- [ ] Feature
- [X] Bugfix
- [ ] Build related changes

## How to test or reproduce?

Open the form to create an application, add a component, and input a git repo URL. The advanced git section won't open, no matter the input.

## Browser conformance: 

- [ ] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ImportForm/ComponentSection/SourceSection.tsx
+++ b/src/components/ImportForm/ComponentSection/SourceSection.tsx
@@ -12,7 +12,7 @@ import GitOptions from './GitOptions';
 
 export const SourceSection = () => {
   const [, { touched, error }] = useField('source.git.url');
-  const [isGitAdvancedOpen, setGitAdvancedOpen] = React.useState<boolean>(false);
+  const [isGitAdvancedOpen, setGitAdvancedOpen] = React.useState<boolean>(true);
   const { touched: touchedValues, setFieldValue } = useFormikContext<ImportFormValues>();
   const validated = touched
     ? touched && !error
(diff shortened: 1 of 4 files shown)


testcara:
Do we need to get UX approval for this change?

sahil143:
> @sahil143 the advanced git options section only has optional/derived values, so I think it's fine to have it closed by default. However, if the git provider is neither Gitlab nor Github, the `Git provider annotation` doesn't get set to anything. Is that okay, or should it be set to `others`? Both gitlab and github will automatically set this field.

I think it would be better to open it by default, Allowing user to review and change any values

marcin-michal:
> I think it would be better to open it by default, Allowing user to review and change any values. WDYT?

Having it open by default makes sense too. I don't really see the reason even to have it collapsible then, as I don't expect many people to collapse it, but I'd say it's almost a non-issue.

rrosatti (approved):
LGTM

src/components/ImportForm/ComponentSection/SourceSection.tsx
```diff
@@ -72,12 +69,8 @@ export const SourceSection = () => {
...
         onChange={handleChange}
       />
-      {validated === ValidatedOptions.success ? (
-        <SwitchField name="isPrivateRepo" label="Should the image produced be private?" />
-      ) : null}
-      {validated === ValidatedOptions.success ? (
-        <GitOptions isGitAdvancedOpen={isGitAdvancedOpen} setGitAdvancedOpen={setGitAdvancedOpen} />
-      ) : null}
+      <SwitchField name="isPrivateRepo" label="Should the image produced be private?" />
+      <GitOptions isGitAdvancedOpen={isGitAdvancedOpen} setGitAdvancedOpen={setGitAdvancedOpen} />
```
sahil143, line 73: These fields should be only visible after a valid git url is entered.

sahil143 (changes requested):

testcara (dismissed):
lgtm

rrosatti (dismissed):
LGTM

rakshett:
LGTM

JoaoPedroPP (approved):
LGTM

UI is working as expected

src/components/ImportForm/ComponentSection/SourceSection.tsx
marcin-michal: done

sahil143 (approved):

StanislavJochman (approved):
LGTM

rrosatti (approved):
LGTM!

PR #166, opened 2025-03-21
feat(KFLUXUI-377): enter, tab and losing focus would input username

testcara:
## Fixes 
[KFLUXUI-377](https://issues.redhat.com/browse/KFLUXUI-377)

## Description
Previously, when users grant user access, they input the username and the username would become inputted username chip automatically. During the SpaceBindingRequest migration, we did one improvement to ensure users need to enter 'username'.  It changes the old behavior then users would feel bad for entering.

With the current patch, users do not have to enter usernames,  losing focus and tabbing out also work. 

## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

We tried to select roles and then input usernames and vise verse. 
and more, entering, tabbing and losing focus all works well.

https://github.com/user-attachments/assets/5fe2e062-2215-49d5-8778-8431ab7c5e23

## How to test or reproduce?
Grant users with different ways('losing focus', 'tabbing out', 'entering') of inputting username

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/UserAccess/UserAccessForm/UserAccessForm.tsx
+++ b/src/components/UserAccess/UserAccessForm/UserAccessForm.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Form, PageSection, PageSectionVariants } from '@patternfly/react-core';
-import { FormikProps } from 'formik';
+import { FormikProps, useFormikContext } from 'formik';
 import isEmpty from 'lodash/isEmpty';
 import PageLayout from '../../../components/PageLayout/PageLayout';
 import { FormFooter } from '../../../shared';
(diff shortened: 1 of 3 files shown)


sahil143 (changes requested):
1. After adding one username submit button is disabled, it should be enabled
2. When submit button is disabled pressing enter submits the form.
3. There is no information on the page indicating that pressing `enter` would allow user to enter multiple user names


https://github.com/user-attachments/assets/c49c4db5-abe5-4d9d-aa85-7d6e6e477b64

StanislavJochman (dismissed):
LGTM

testcara:
> 1. After adding one username submit button is disabled, it should be enabled
> 
>     2. When submit button is disabled pressing enter submits the form.
> 
>     3. There is no information on the page indicating that pressing `enter` would allow user to enter multiple user names
> 
> 
> Kapture.2025-03-26.at.18.22.04.mp4
For 1:
It is quite interesting, if we input username first and choose roles later, all things work.
When we choose roles first and input username later, the button does not work well. :-).
Let me try to improve a bit.
For 2: 
I guess if we resolve 1, the 2 would be fixed.
For 3:
For we both support 'tab' & 'enter', as recommended by Pankaj, we do not need to explain it.  if needed, let us involve UX again. :-). Both ideas are ok for me.

testcara:
FYI. The error of UserAccessForm does not affect the username section errors show-up. 
When there is any real error, the error returned looks like {"usernames":"Must be 2 to 45 characters long and can only contain letters from a to Z, numbers from 0 to 9, underscores( _ ), hyphens( - ), periods( . ), or @ symbol."} which does not meet the condition ```errors?.usernames === ''```. 
That is to say, the below codes would just fix the fake error reported by formik.
```
 const customErrors =
    errors?.usernames === '' && values?.usernames.length > 0 ? undefined : errors;
```

sahil143 (approved):

PR #174, opened 2025-03-26
feat(KFLUXUI-7307): add format to kebab case for name in github url

rakshett:
## Fixes 

https://issues.redhat.com/browse/KONFLUX-7307

## Description

This PR enhances the SourceSection component by automating component name formatting to kebab-case and improving Git provider detection logic. It also adds unit tests to validate these changes

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

image

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


src/components/ImportForm/ComponentSection/SourceSection.tsx
```diff
@@ -19,15 +18,14 @@ export const SourceSection = () => {
...
-    return uniqueName;
-  }
+  const formatToKebabCase = (name: string): string =>
+    name
+      .replace(/_/g, '-')
+      .replace(/([a-z])([A-Z])/g, '$1-$2')
+      .replace(/(\d+)/g, '-$1')
+      .toLowerCase()
+      .replace(/-+/g, '-')
+      .replace(/^-|-$/g, '');
```
sahil143, line 28: create one regex that takes care of all the patterns

src/components/ImportForm/ComponentSection/SourceSection.tsx
```diff
@@ -19,15 +18,14 @@ export const SourceSection = () => {
...
     : ValidatedOptions.default;
 
-  function generateRandomString(): string {
-    let uniqueName: string;
-    do {
-      uniqueName = uuidv4()
-        .replace(/[^a-z0-9-]/g, '')
-        .substring(0, 5);
-    } while (!/^[a-z][a-z0-9-]*[a-z0-9]$/.test(uniqueName));
-    return uniqueName;
```
sahil143, line 29: we need unique id. it should be appended to the extracted component name.

rakshett: Updated

src/components/ImportForm/ComponentSection/SourceSection.tsx
rakshett: Updated for the first few patterns. unable to update all as this is not supporting

StanislavJochman (approved):
LGTM

src/components/ImportForm/ComponentSection/SourceSection.tsx
sahil143: What is not supporting?

src/components/ImportForm/ComponentSection/SourceSection.tsx
```diff
@@ -47,18 +58,29 @@ export const SourceSection = () => {
...
         } catch {
+          name = '';
           await setFieldValue('gitURLAnnotation', '');
         }
+
         if (!touchedValues.componentName) {
-          await setFieldValue('componentName', generateRandomString());
+          let formattedName = formatToKebabCase(name);
+          const namespaceMatch = namespaces.some(
+            (namespace) => namespace.metadata.name === formattedName,
```
sahil143, line 74: why are we matching it with namespace name?

sahil143 (changes requested):

src/components/ImportForm/ComponentSection/SourceSection.tsx
rakshett: We need to find whether the formatted namespace already exists in the current namespaces list incase if it exists we are appending unique characters at the end

src/components/ImportForm/ComponentSection/SourceSection.tsx
rakshett: Tried with different patterns of regex but is not supporting for our usecase hence we are adding this

rakshett: Updated

src/components/ImportForm/ComponentSection/SourceSection.tsx
```diff
@@ -29,6 +33,21 @@ export const SourceSection = () => {
     return uniqueName;
   }
 
+  const formatToKebabCase = (name: string): string =>
```
rrosatti, line 36: small suggestion: we can move `formatToKebabCase` outside the `SourceSection` component, so we avoid re-defining the function on each re-render :)

or maybe even move to some "utils" file, in case the function could be re-used somewhere else

src/components/ImportForm/ComponentSection/SourceSection.tsx
```diff
@@ -29,6 +33,21 @@ export const SourceSection = () => {
...
+    name.replace(
+      /([a-z])([A-Z])|_|(\d+)|(^-+|-+$)|(-+)/g,
```
rrosatti, line 38: also, I took a quick look and it seems the `lodash` lib has a `kebabCase` function (https://lodash.com/docs/#kebabCase) - maybe it could be used here to simplify the code? 🙂

rakshett: @rrosatti Initially we tried with lodash itself but it didn't fix the below scenario
Using lodash.kebabCase would not handle all the specific cases in the provided regex, such as adding hyphens before digits or preserving certain uppercase letters hence,  custom implementation is tailored to meet specific formatting requirements that lodash.kebabCase does not natively support. :)

src/components/ImportForm/ComponentSection/SourceSection.tsx
rakshett: @rrosatti We don't have any usecases where this is required in other file. We shall add when required

src/components/ImportForm/ComponentSection/SourceSection.tsx
sahil143: But why namespace? this section has nothing to do namespace. Also, we don't need to compare it with components. It should be as simple as

```
<git-repo-name>-<uuid()>
```

src/components/ImportForm/ComponentSection/SourceSection.tsx
sahil143: Can you please give examples of cases that are not supported by lodash but are needed by component section?

src/components/ImportForm/ComponentSection/SourceSection.tsx
```diff
@@ -4,13 +4,17 @@ import { useField, useFormikContext } from 'formik';
...
 import { detectGitType, GitProvider } from '../../../shared/utils/git-utils';
 import { GIT_PROVIDER_ANNOTATION_VALUE } from '../../../utils/component-utils';
 import { ImportFormValues } from '../type';
 import GitOptions from './GitOptions';
 
-export const SourceSection = () => {
+export const SourceSection: React.FC<{ applicationName: string }> = ({ applicationName }) => {
   const [, { touched, error }] = useField('source.git.url');
+  const namespace = useNamespace();
+  const [components] = useComponents(namespace, applicationName);
```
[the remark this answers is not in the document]
rakshett: Please advise on how to get the components without fetching. We need to check the name of the component if it exists then we will be adding the uniquely generated text.
@sahil143

src/components/ImportForm/ComponentSection/SourceSection.tsx
rakshett: We are using namespace as the helper text defines `Must be unique within tenant namespace` hence the namespace is required and useComponents is required to fetch the list

src/components/ImportForm/ComponentSection/SourceSection.tsx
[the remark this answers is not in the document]
sahil143: you dont need to fetch the components. Generated uuid are unique by default so if you append a uuid to a name it will be unique in the namespace.

src/components/ImportForm/ComponentSection/SourceSection.tsx
rakshett: Updated using lodash kebabcase

src/components/ImportForm/ComponentSection/SourceSection.tsx
rakshett: Updated using lodash kebabcase as this function is no longer required

src/components/ImportForm/ComponentSection/__tests__/ComponentSection.spec.tsx
```diff
@@ -75,4 +75,21 @@ describe('ComponentSection', () => {
...
+    });
+    const user = userEvent.setup();
+    const source = screen.getByPlaceholderText('Enter your source');
+
+    await user.type(source, 'https://github.com/ExampleRepo123.git');
+    await user.tab();
+    await waitFor(() => {
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+      expect((screen.getByTestId('component-name') as HTMLInputElement).value).toContain(
+        'example-repo',
```
rrosatti, line 91: shouldn't the expected result be something like `example-repo-123-X`? 🤔  - where X is the result of `generateRandomString()`

src/components/ImportForm/ComponentSection/__tests__/ComponentSection.spec.tsx
```diff
@@ -75,4 +75,21 @@ describe('ComponentSection', () => {
...
+    formikRenderer(<ComponentSection />, {
+      source: { git: { url: '' } },
+    });
+    const user = userEvent.setup();
+    const source = screen.getByPlaceholderText('Enter your source');
+
+    await user.type(source, 'https://github.com/ExampleRepo123.git');
+    await user.tab();
+    await waitFor(() => {
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
sahil143, line 89: Why do you need to disable eslint here?

sahil143 (changes requested):

src/components/ImportForm/ComponentSection/__tests__/ComponentSection.spec.tsx
rakshett: @rrosatti Since this will generate the unique characters on every test so, we are only checking the initial characters with using toContain method

src/components/ImportForm/ComponentSection/__tests__/ComponentSection.spec.tsx
rakshett: The eslint is not letting to push the commit.

rrosatti: have you tried using the `.getAttribute()` function?

it would be something like this: 

```typescript
expect((screen.getByTestId('component-name').getAttribute('value')).toContain(...
```

if it works, then you won't need the type assertion :)

rakshett: I tried the same, But it's not working.

sahil143 (approved):

PR #185, opened 2025-04-01
fix: use `konflux-<role>-<username>-user-actions` naming convertion for role binding names

sahil143:
## Description

This PR uses the naming convention for role binding names

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes

no visual changes

## How to test or reproduce?

Test User access:
  - creation flow
  - edit flow
  - list view
  - revoke flow

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/UserAccess/UserAccessForm/__tests__/form-utils.spec.ts
+++ b/src/components/UserAccess/UserAccessForm/__tests__/form-utils.spec.ts
@@ -1,11 +1,9 @@
 import '@testing-library/jest-dom';
 import { defaultKonfluxRoleMap } from '../../../../__data__/role-data';
 import { mockRoleBinding } from '../../../../__data__/rolebinding-data';
-import { K8sQueryDeleteResource } from '../../../../k8s';
-import { RoleBindingModel } from '../../../../models';
 import { NamespaceRole } from '../../../../types';
 import { createK8sUtilMock } from '../../../../utils/test-utils';
-import { createRBs, editRB, deleteRB } from '../form-utils';
+import { createRBs, editRB } from '../form-utils';
 
 const k8sCreateMock = createK8sUtilMock('K8sQueryCreateResource');
 const k8sGetMock = createK8sUtilMock('K8sQueryListResourceItems');
@@ -31,9 +29,9 @@ describe('createRBs', () => {
     // Because althrough we use expect.objectContaining(), removing some items
     // of roleRef still make test fail.
     const expectedResult = [
-      { 'uuser1---123': '-Uuser1@!#123' },
-      { user2: 'user2-$~+.' },
-      { user3: 'user3_//-' },
+      { 'konflux-maintainer-uuser1---123-user-actions': '-Uuser1@!#123' },
+      { 'konflux-maintainer-user2-user-actions': 'user2-$~+.' },
+      { 'konflux-maintainer-user3-user-actions': 'user3_//-' },
     ].map((userObj) => {
       const [sanitizedName, rawName] = Object.entries(userObj)[0]; // 解构键值对，获取键和值
 
@@ -83,18 +81,17 @@ describe('editRBs', () => {
 
     k8sGetMock.mockResolvedValue([mockRoleBinding]);
     k8sDeleteMock.mockImplementation();
-    // There is one bug for K8sQueryDeleteResource. When the dryRun is true, it would
-    // also delete resources. So we test dryRun as true for successfuly deletion.
+
     const result = await editRB(values, mockRoleBinding, true);
 
-    expect(k8sDeleteMock).toHaveBeenCalledTimes(1);
+    expect(k8sDeleteMock).toHaveBeenCalledTimes(0);
     expect(k8sCreateMock).toHaveBeenCalledTimes(1);
     expect(result).toEqual([
       expect.objectContaining({
         ...mockRoleBinding,
         metadata: {
           ...mockRoleBinding.metadata,
-          name: `user1`,
+          name: `konflux-admin-user1-user-actions`,
         },
         subjects: mockRoleBinding.subjects.map((subject) => ({
           ...subject,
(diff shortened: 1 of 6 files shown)


src/components/UserAccess/UserAccessForm/form-utils.ts
```diff
@@ -97,14 +97,7 @@ export const deleteRB = async (roleBinding: RoleBinding, dryRun?: boolean): Prom
...
   };
 
-  // K8sQueryDeleteResource enjoys the k8sDeleteResource to delete resources.
-  // However, based on the design of k8sDeleteResource, 'dryRun' just affect the
-  // query parameter but would not affect commonFetchJSON.delete.
-  // That is to say, with 'true' dryRun, the resource would be also deleted.
-  // If so, we have to skip the redeletion when dryRun is false.
-  // To do: let us improve the function when the K8sQueryDeleteResource
-  // works well with dryRun.
-  dryRun ? await K8sQueryDeleteResource(queryOptions) : void 0;
```
sahil143, line 107: If dry run is deleting resources, the we should not call delete resource method in dry run phase.

src/components/UserAccess/UserAccessForm/__tests__/form-utils.spec.ts
```diff
@@ -114,46 +111,10 @@ describe('editRBs', () => {
...
     k8sCreateMock.mockRejectedValueOnce(new Error('Create failed'));
 
-    await expect(editRB(values, mockRoleBinding)).rejects.toThrow('Create failed');
-
+    await expect(editRB(values, mockRoleBinding, true)).rejects.toThrow('Create failed');
+    // error should be caught in creation during dry run
     expect(k8sCreateMock).toHaveBeenCalledTimes(1);
-    // If users cannot be created successfully, we do not delete existing roles.
-    // In this way, it looks like users does not edit roles successfuly.
+    // delete should not be called during dry run
```
StanislavJochman, line 117: does this apply above as well?

StanislavJochman (approved):
LGTM

src/components/UserAccess/UserAccessForm/form-utils.ts
```diff
@@ -56,7 +56,7 @@ export const createRBs = async (
     metadata: {
       // To sanitize the username and ensure every user just has one role
       // in the namespace.
-      name: `${sanitizeUsername(username)}`,
+      name: `konflux-${role.toLowerCase()}-${sanitizeUsername(username)}-user-actions`,
```
abhinandan13jan, line 59: Should we store this `-user-actions` in const like USER_ACTION_SUFFIX?

src/components/UserAccess/UserAccessForm/__tests__/form-utils.spec.ts
sahil143: apply above? where?

src/components/UserAccess/UserAccessForm/__tests__/form-utils.spec.ts
```diff
@@ -83,18 +81,17 @@ describe('editRBs', () => {
 
     k8sGetMock.mockResolvedValue([mockRoleBinding]);
     k8sDeleteMock.mockImplementation();
-    // There is one bug for K8sQueryDeleteResource. When the dryRun is true, it would
-    // also delete resources. So we test dryRun as true for successfuly deletion.
+
     const result = await editRB(values, mockRoleBinding, true);
 
-    expect(k8sDeleteMock).toHaveBeenCalledTimes(1);
+    expect(k8sDeleteMock).toHaveBeenCalledTimes(0);
```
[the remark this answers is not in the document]
sahil143: yes, because during dry run delete resource method should not be called.

src/k8s/query/fetch.ts
```diff
@@ -70,12 +70,20 @@ export const K8sQueryDeleteResource = <TResource extends K8sResourceCommon>(
   requestInit: K8sResourceDeleteOptions,
 ) => {
   return k8sDeleteResource<TResource>(requestInit).finally(() => {
-    !requestInit.queryOptions?.queryParams?.dryRun &&
+    if (!requestInit.queryOptions?.queryParams?.dryRun) {
```
testcara, line 73: how about we add some comments here? so with the change, the k8sDeleteResource just delete resources when dryRun is false, right?

src/components/UserAccess/UserAccessForm/form-utils.ts
```diff
@@ -124,14 +117,7 @@ export const editRB = async (
...
 
-  // We need to delete the role binding before we create the new one.
-  const existingRBs = await getRBs(roleBinding.metadata.namespace);
-
-  const roleBindingExists = existingRBs.some(
-    (binding) => binding.metadata.name === roleBinding.metadata.name,
-  );
-
-  if (roleBindingExists) {
+  if (!dryRun) {
```
testcara, line 120: I saw in the k8sDeleteResource, we have improved the condition.
Whether we still need to add condition here? what will happen if we do not have the condition?

src/components/UserAccess/UserAccessForm/form-utils.ts
testcara: How we avoid 'one user has multiple accounts'? Let the doc team improve this part>

src/k8s/query/fetch.ts
sahil143: this is not for deleting resource. This just removes the query from cache so it doesn't refetch data internally.

src/components/UserAccess/UserAccessForm/form-utils.ts
sahil143: K8sDeleteResource just removed the condition, this condition should be handled by consumer of deleteRB.
There were two things, 
1. if we allowed resource to be deleted during dryrun phase. It tried to refetch the same resource which caused 404 error on API requests
2.DryRun should allow us to catch any errors before hand, so that if user isn't able to create new RB then previous RB still remains and target username doesn't loose previous permissions to the namespace.

testcara (approved):

sahil143:
I tested locally, different e2e tests are failing with each run and PR checks failure doesn't seem to be related with changes. Merging PR. Any changes required for e2e test failure can be covered in follow up PR as this PR is blocking UI release.

PR #190, opened 2025-04-07
fix(KONFLUX-5796): the nudging should be enabled for app has a single component

testcara:
## Fixes 
[KONFLUX-5796](https://issues.redhat.com/browse/KONFLUX-5796)

## Description
Without the patch, when the app has a single component, the nudge action is disabled.
With the patch, when the app has a single component and there are one more components in the namespace, the nudging action would be enabled.

Note:
When there is no component, or there is one single component in the namespace, the nudging action would be kept as disabled.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Screenshot 2025-04-07 at 18 25 31

## How to test or reproduce?
1. create one app with one component and try to update the component relationship
Result:
When there is no other components in the namespace, we cannot update the relationship
When there is one more components in the namespace, we can update the relationship
2. create one app without component
Result:
we cannot update the relationship

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/ApplicationDetails/__data__/WorkflowTestUtils.ts
+++ b/src/components/ApplicationDetails/__data__/WorkflowTestUtils.ts
@@ -1,4 +1,4 @@
-import { useComponents, useSortedComponents } from '../../../hooks/useComponents';
+import { useAllComponents, useComponents, useSortedComponents } from '../../../hooks/useComponents';
 import { useIntegrationTestScenarios } from '../../../hooks/useIntegrationTestScenarios';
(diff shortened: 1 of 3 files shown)


StanislavJochman (approved):
LGTM

rrosatti (approved):
LGTM

JoaoPedroPP (approved):
LGTM

The UI is working as expected

PR #202, opened 2025-04-11
KONFLUX-7750: Fix sorting issue on Applications list by name

rrosatti:
## Fixes 

This fixes the issue https://issues.redhat.com/browse/KONFLUX-7750

## Description

This PR implements sorting by name on the "Applications" page. While working on this feature, I noticed a similar behavior on the "Releases" page. So as a result,  I refactored the logic for creating table headers into a more generic utility function called `createTableHeaders`, which is now used on both pages.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

**Applications page:**

applications-with-sorting

**Releases page (to show that nothing changes and it still works as expected):**

releases-refactor-no-changes

## How to test or reproduce?

1. Make sure you have a few applications
2. Go to "Applications" page and now you'll notice that it's possible to sort the name column
3. ☕ 

Additional testing:

1. Since I also updated `ReleasesListHeader` to use the `createTableHeaders` utilities, you can also check the "Releases" page to see that it continues working as expected 

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/Applications/ApplicationListHeader.ts
+++ b/src/components/Applications/ApplicationListHeader.ts
@@ -1,3 +1,5 @@
+import { createTableHeaders } from '../../shared/components/table/utils';
+
 export const applicationTableColumnClasses = {
   name: 'pf-m-width-35',
   components: 'pf-m-width-30',
@@ -5,25 +7,15 @@ export const applicationTableColumnClasses = {
   kebab: 'pf-v5-c-table__action',
 };
 
-export const ApplicationListHeader = () => {
-  return [
-    {
-      title: 'Name',
-      props: { className: applicationTableColumnClasses.name },
-    },
-    {
-      title: 'Components',
-      props: { className: applicationTableColumnClasses.components },
-    },
-    {
-      title: 'Last deploy',
-      props: { className: applicationTableColumnClasses.lastDeploy },
-    },
-    {
-      title: ' ',
-      props: {
-        className: applicationTableColumnClasses.kebab,
-      },
-    },
-  ];
-};
+export const enum SortableHeaders {
+  name,
+}
+
+const appColumns = [
+  { title: 'Name', className: applicationTableColumnClasses.name, sortable: true },
+  { title: 'Components', className: applicationTableColumnClasses.components },
+  { title: 'Last deploy', className: applicationTableColumnClasses.lastDeploy },
+  { title: ' ', className: applicationTableColumnClasses.kebab },
+];
+
+export default createTableHeaders(appColumns);
--- a/src/components/Applications/ApplicationListView.tsx
+++ b/src/components/Applications/ApplicationListView.tsx
@@ -11,10 +11,12 @@ import {
   ToolbarContent,
   ToolbarItem,
 } from '@patternfly/react-core';
+import { SortByDirection } from '@patternfly/react-table';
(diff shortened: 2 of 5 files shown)


src/utils/table-list-utils.ts
```diff
@@ -0,0 +1,32 @@
+import { SortByDirection, ThProps } from '@patternfly/react-table';
+import { HeaderFunc } from '~/shared/components/table/Table';
+
+export type ColumnConfig = {
+  title: string;
+  className: string;
+  sortable?: boolean;
+};
+
```
testcara, line 9: great utils

testcara (commented):
Here is the original slack thread to discuss the problem:
https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1744209857716789

It seems people prefer to get the sorted application without clicking 'sort' directly.
I guess it would be helpful to double confirm with the design with arewm.

src/utils/table-list-utils.ts
```diff
@@ -0,0 +1,32 @@
+import { SortByDirection, ThProps } from '@patternfly/react-table';
+import { HeaderFunc } from '~/shared/components/table/Table';
+
+export type ColumnConfig = {
+  title: string;
+  className: string;
+  sortable?: boolean;
```
testcara, line 7: How about we add style now to ensure it is more flexible?

rrosatti:
> Here is the original slack thread to discuss the problem: https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1744209857716789
> 
> It seems people prefer to get the sorted application without clicking 'sort' directly. I guess it would be helpful to double confirm with the design with arewm.

thanks for the link! 

from the code I'm adding in this PR, it'll be sorting by ascending order by default (so the user don't need to click sort directly).
I also I left a comment in the ticket: https://issues.redhat.com/browse/KONFLUX-7750 regarding the possibility of having an additional column for the "creation timestamp" info. Do you think we should get the opinion of someone else as well? 🤔

src/utils/table-list-utils.ts
rrosatti: that's a good idea, I'll update the code!

testcara (dismissed):

janaki29 (dismissed):
lgtm

src/utils/table-list-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+  activeDirection: SortByDirection,
+  onSort: ThProps['sort']['onSort'],
+) => HeaderFunc) => {
+  return (activeIndex, activeDirection, onSort) => () => {
+    const getSortParams = (columnIndex: number) => ({
+      columnIndex,
+      sortBy: { index: activeIndex, direction: activeDirection },
+      onSort,
+    });
+
```
sahil143, line 24: These are specific to Table component. Can we move these to shared folder with the Table component itself.

sahil143 (changes requested):

src/utils/table-list-utils.ts
rrosatti: updated! :)

testcara (approved):
lgtm

JoaoPedroPP (approved):
LGTM

UI is working as expected

PR #205, opened 2025-04-14
KFLUXUI-232: [Commit Details - PipelineRun Tab] Update to also show "component-less" PipelineRuns

rrosatti:
## Fixes 

This fixes the issue https://issues.redhat.com/browse/KFLUXUI-232

## Description

This PR is adding the changes to allow displaying in the "Commit Details - PipelineRun Tab", the `PipelineRun`s that doesn't have a component directly related to it.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Before the fix (notice that the PipelineRun `tenant-bdvhf` is not displayed):

commit-pipelinerun-without-component-issue

With the fix:

commit-pipelinerun-without-component-fix

## How to test or reproduce?

1. Find some `PipelineRun` that doesn't have a component
2. Open the Commit Details - PipelineRuns Tab
3. The "component-less" `PipelineRun` should be listed there
4. ☕ 

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/hooks/usePipelineRuns.ts
+++ b/src/hooks/usePipelineRuns.ts
@@ -247,6 +247,7 @@ export const usePipelineRunsForCommit = (
   applicationName: string,
   commit: string,
   limit?: number,
+  filterByComponents = true,
 ): [PipelineRunKind[], boolean, unknown, GetNextPage, NextPageProps] => {
   const [components, componentsLoaded] = useComponents(namespace, applicationName);
   const [application, applicationLoaded] = useApplication(namespace, applicationName);
@@ -286,7 +287,9 @@ export const usePipelineRunsForCommit = (
     return [
       pipelineRuns
         .filter((plr) =>
-          componentNames.includes(plr.metadata?.labels?.[PipelineRunLabel.COMPONENT]),
+          filterByComponents
+            ? componentNames.includes(plr.metadata?.labels?.[PipelineRunLabel.COMPONENT])
+            : true,
         )
         .filter((plr) => plr.kind === PipelineRunGroupVersionKind.kind)
         .slice(0, limit ? limit : undefined),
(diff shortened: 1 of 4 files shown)


src/utils/commits-utils.ts
```diff
@@ -14,7 +14,8 @@ export const statuses = [
 export const getCommitSha = (pipelineRun: PipelineRunKind) =>
   pipelineRun?.metadata.labels?.[PipelineRunLabel.COMMIT_LABEL] ||
   pipelineRun?.metadata.labels?.[PipelineRunLabel.TEST_SERVICE_COMMIT] ||
-  pipelineRun?.metadata.annotations?.[PipelineRunLabel.COMMIT_ANNOTATION];
+  pipelineRun?.metadata.annotations?.[PipelineRunLabel.COMMIT_ANNOTATION] ||
+  pipelineRun?.metadata?.annotations?.[PipelineRunLabel.TEST_SERVICE_COMMIT];
```
rrosatti, line 18: this is also part of the fix, because I noticed that `pac.test.appstudio.openshift.io/sha` could be part of the `annotations` object instead of `labels`:

pipelinerun-commit-sha-in-annotation

at least it happened in the scenario I was trying to reproduce within my namespace :)

src/hooks/usePipelineRuns.ts
```diff
@@ -247,6 +247,7 @@ export const usePipelineRunsForCommit = (
   applicationName: string,
   commit: string,
   limit?: number,
+  filterByComponents = true,
```
rrosatti, line 250: I noticed the `usePipelineRunsForCommit` hook is being used in a few places, so I decided to go with this approach of adding a new param instead of removing the filtering altogether

but, let me know if we should handle it differently :)

janaki29 (approved):
lgtm

testcara:
how about we add some tests for the patch?

testcara (commented):
The codes look good to me but it would be better if some tests could be added.

rrosatti:
> The codes look good to me but it would be better if some tests could be added.

updated! :)

StanislavJochman (approved):
LGTM

StanislavJochman (approved):

PR #211, opened 2025-04-17
KFLUXUI-230: Add support for "re-running" a Release

rrosatti:
## Fixes 

This fixes the issue https://issues.redhat.com/browse/KFLUXUI-230.

## Description

In this PR we’re introducing logic to allow a `Release` to be “re-run” or “re-triggered”.
Behind the scenes, this doesn’t actually re-run the same resource — instead, it creates a new `Release` based on the existing one.
It kinda simulates a "rerun" by cloning the original release, generating a new name with a random suffix, and updating relevant info (e.g. labels)

The implementation follows the approach described here https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/release-service/faq.md?ref_type=heads#how-do-i-retrigger-a-release, which explains how to "rerun"/"re-trigger" a release via CLI 🙂 

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

The screen recording below shows the "re-run" action in action 😄

https://github.com/user-attachments/assets/0c91223f-901e-44ab-884c-8b6ea4bf64fe

## How to test or reproduce?

1. Go to `Releases` page
2. Notice that by the end of the "Release Row", there's an `ActionMenu`
3. Click on it and select the "Re-run release" option
4. Notice that a new release was created and it's in progress now
5. ☕ 

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/consts/release.ts
+++ b/src/consts/release.ts
@@ -1,0 +1,4 @@
+export enum ReleaseLabel {
+  AUTOMATED = 'release.appstudio.openshift.io/automated',
+  AUTHOR = 'release.appstudio.openshift.io/author',
+}
--- a/src/utils/release-actions.ts
+++ b/src/utils/release-actions.ts
@@ -1,0 +1,39 @@
+import { ReleaseLabel } from '../consts/release';
+import { K8sQueryCreateResource } from '../k8s';
+import { ReleaseModel } from '../models';
+import { ReleaseKind } from '../types';
+import { generateNewReleaseName } from './release-utils';
+
+export const releaseRerun = (release: ReleaseKind, username: string) => {
+  const newName = generateNewReleaseName(release.metadata?.name ?? '');
+
+  const newRelease: ReleaseKind = {
+    ...release,
+    metadata: {
+      ...release.metadata,
+      creationTimestamp: undefined,
+      finalizers: undefined,
+      generateName: `${newName}`,
+      generation: undefined,
+      ownerReferences: undefined,
+      resourceVersion: undefined,
+      uid: undefined,
+      name: undefined,
+      labels: {
+        ...release.metadata.labels,
+        [ReleaseLabel.AUTOMATED]: 'false',
+        [ReleaseLabel.AUTHOR]: username,
+      },
+    },
+    status: undefined,
+  };
+
+  return K8sQueryCreateResource({
+    model: ReleaseModel,
+    queryOptions: {
+      name: release.metadata.name,
+      ns: release.metadata.namespace,
+    },
+    resource: newRelease,
+  });
+};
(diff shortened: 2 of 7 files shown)


src/consts/release.ts
```diff
@@ -0,0 +1,5 @@
+export enum ReleaseLabel {
+  APPLICATION = 'appstudio.openshift.io/application',
```
testcara, line 2: I saw same const defined:
 src/consts/pipelinerun.ts:  APPLICATION = 'appstudio.openshift.io/application',

and the const has been enjoyed by other components. so i am considering how about we enjoy the existing const for this one.

abhinandan13jan: You can use the Application label from PipelineRunLabels in `src/consts/pipelinerun.ts`

src/utils/release-utils.ts
```diff
@@ -19,3 +19,26 @@ export const getTenantPipelineRunFromRelease = (release: ReleaseKind): string =>
...
   return release.status?.finalProcessing?.pipelineRun;
 };
+
+// Example:
+// Math.random().toString(36)               // "0.f5b27xpmc"
+// Math.random().toString(36).slice(2, 7)   // "f5b27"
+const generateSuffix = (length = 5): string =>
+  Math.random()
+    .toString(36)
+    .slice(2, 2 + length);
```
abhinandan13jan, line 29: Do we need this specifically? We can set `metadata.generateName = prefix` while creating the resource, that should do it for you in most cases

src/utils/release-actions.ts
```diff
@@ -0,0 +1,39 @@
...
+export const releaseRerun = (release: ReleaseKind, username: string) => {
+  const newName = generateNewReleaseName(release.metadata?.name ?? '');
+
+  const newRelease: ReleaseKind = {
+    ...release,
+    metadata: {
+      ...release.metadata,
+      creationTimestamp: undefined,
+      finalizers: undefined,
+      generateName: undefined,
```
abhinandan13jan, line 16: `metadata.generaeName` can be used for creating unique names with suffix. Not sure if there is some particular use case which we want to address

src/utils/release-utils.ts
rrosatti: hmm, good question - I'll give it a try to update the `metadata.generateName` instead of `name` (without this "suffix" logic) and check if it works :)

testcara (approved):
lgtm

marcin-michal (approved):
LGTM

janaki29 (approved):

rakshett (approved):
LGTM

StanislavJochman (approved):
LGTM

PR #226, opened 2025-05-07
feat(FeatureFlags): [KFLUXUI-451] add feature flags capability

sahil143:
## Fixes 

https://issues.redhat.com/browse/KFLUXUI-451

## Description

Please go through the added readme for feature flags

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes

image
image

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/docs/feature-flags.md
+++ b/docs/feature-flags.md
@@ -1,0 +1,116 @@
+# 🚦 Feature Flagging — Tech Docs
+
+This document explains how our **client‑side** feature‐flag system works, how flags are stored, and how to add or consume a flag in daily development.
+
+---
+
+## 1. What lives where?
+
+```
+src/
+  feature-flags/
+    flags.ts     # canonical list, default states, metadata
+    store.ts     # runtime state resolver (localStorage + URL)
+    panel.tsx    # React component to disable and enable a feature flag
+    hooks.ts     # React hooks and component to disable code
+```
+
+*There is **no** backend call. Everything happens in the browser.*
+
+---
+
+## 2. Persistence & update flow
+
+| Step | Source           | Example                              | Notes                                        |
+| ---- | ---------------- | ------------------------------------ | -------------------------------------------- |
+| 1    | **Default**      | `defaultEnabled` in `flags.ts`       | Baseline value shipped in the bundle.        |
+| 2    | **localStorage** | `__ff_overrides__ => { "foo": true }` | Persists your choices across reloads.        |
+| 3    | **URL params**   | `?ff=foo,bar` or `?ff_foo=false`    | Highest priority; great for shareable links. |
+
+**Setting a flag with our API does two things:**
+
+1. Writes to localStorage so the change sticks.
+2. Patches the current URL (`history.replaceState`) so the address bar shows the new state without a full reload.
+
+---
+
+### URL grammar cheat‑sheet
+
+* `?ff=flagA,flagB` → turn *on* multiple flags quickly.
+* `?ff_flagA=true|false` → explicit on/off for one flag.
+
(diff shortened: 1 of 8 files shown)


StanislavJochman (dismissed):
LGTM just out of curiosity because I am not familiar with local storage. Is there a way to force its recreation? For example, let’s say you store some flags in it, and then we push a new version of the UI that adds or removes certain items. What happens in that case? Does local storage update automatically, or does the user need to clear it manually?

rrosatti (approved):
LGTM!

docs/feature-flags.md
```diff
@@ -0,0 +1,116 @@
...
+
+1. Writes to localStorage so the change sticks.
+2. Patches the current URL (`history.replaceState`) so the address bar shows the new state without a full reload.
+
+---
+
+### URL grammar cheat‑sheet
+
+* `?ff=flagA,flagB` → turn *on* multiple flags quickly.
+* `?ff_flagA=true|false` → explicit on/off for one flag.
```
testcara, line 40: whether the flagA/flagB here is the key of feature flag, like kubearchive and buildServiceAccount? 

It seems we could enable feature on any URL of UI? 
I tried the  https://localhost:8080/ns?ff=buildServiceAccount locally but nothing happens. 
What is the right url i could see the panel?

docs/feature-flags.md
```diff
@@ -0,0 +1,116 @@
...
+
+| Step | Source           | Example                              | Notes                                        |
+| ---- | ---------------- | ------------------------------------ | -------------------------------------------- |
+| 1    | **Default**      | `defaultEnabled` in `flags.ts`       | Baseline value shipped in the bundle.        |
+| 2    | **localStorage** | `__ff_overrides__ => { "foo": true }` | Persists your choices across reloads.        |
+| 3    | **URL params**   | `?ff=foo,bar` or `?ff_foo=false`    | Highest priority; great for shareable links. |
+
+**Setting a flag with our API does two things:**
+
+1. Writes to localStorage so the change sticks.
```
testcara, line 32: I was considering the useContext + localstorage. but url + localstorage is also good choice.

testcara:
> LGTM just out of curiosity because I am not familiar with local storage. Is there a way to force its recreation? For example, let’s say you store some flags in it, and then we push a new version of the UI that adds or removes certain items. What happens in that case? Does local storage update automatically, or does the user need to clear it manually?

you could consider the local storage is a browser-based storage that allows you to persistently save key-value pairs.
When we push a new version of UI,  unless you trigger some UI feature to rewrite/remove your localstorage or clear it manually, it would be always there. 
There is still one potential risk that if your localstorage is full, UI would fail to write it with new settings.

testcara:
This is a great patch. With it, we have no concern to release partial EPIC features!:).

docs/feature-flags.md
sahil143: yes, you can enable features from any of the routes using the use Params. To see the panel click the flask icon in the toolbar.

sahil143: > I tried the https://localhost:8080/ns?ff=buildServiceAccount locally but nothing happens.

Do we have any features on the namespace page related to build service account

testcara: https://localhost:8080/ns/wlin-tenant/secrets/create?ff=buildServiceAccount  does not work for me, either.

testcara (approved):

rrosatti (approved):

StanislavJochman (approved):
LGTM

testcara (approved):
Great patch, thank you.

PR #229, opened 2025-05-09
feat(KFLUXUI-418): add task store & async queue for linking in background

testcara:
## Fixes 
[KFLUXUI-418](https://issues.redhat.com/browse/KFLUXUI-418)

## Description
The patch ensure the linking secrets action would run in background after secret creation.
Task store would track the status of the task.
When error happens, the annotation for linking error would be added to the created secrets.

## Why choose 'async queue + zustand + annotation' but not others
To support I/O heavy tasks, the async queue will not block the main thread.
zustand to ensure the local status would be tracked by an alert to remind the user to keep the window open.
annotation ensures the status can be shared between different users and sessions.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

Recording for  importing secrets as an example:

https://github.com/user-attachments/assets/40d68464-d466-4269-b484-d07eb054f7aa

See the alert and error alert:
Screenshot 2025-06-05 at 21 56 12

## How to test or reproduce?
1. enable the feature flag
2. add secrets with linking components
3. the secret list tracks the status well.

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


rrosatti (dismissed):
LGTM - once the other "dependent PRs" are merged, I'll come back!

I liked the usage of `zustand` here :)

package.json
```diff
@@ -50,7 +49,8 @@
     "react-router-dom": "^6.25.1",
     "sanitize-html": "^2.13.0",
     "showdown": "^2.1.0",
-    "yup": "^0.32.11"
+    "yup": "^0.32.11",
+    "zustand": "^5.0.4"
```
sahil143, line 53: Why did you opt for `zustand` instead of `jotai`?

src/components/Secrets/SecretsListView/SecretsListRowTooltip.scss
```diff
@@ -0,0 +1,15 @@
...
+  --pf-v5-c-tooltip__content--Padding: var(--pf-v5-global--spacer--sm);
+  --pf-v5-c-tooltip__content--BackgroundColor: var(--pf-v5-c-tooltip--BackgroundColor);
+  --pf-v5-c-tooltip__content--Color: var(--pf-v5-c-tooltip--Color);
+}
+
+.pf-v5-c-tooltip__content {
+  white-space: normal;
+  word-break: break-word;
+  text-align: left;
+}
```
sahil143, line 15: Why do you need change default tooltip styles?

src/utils/task-store.ts
```diff
@@ -0,0 +1,42 @@
...
+        set((state) => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          const { [id]: _, ...rest } = state.tasks;
+          return { tasks: rest };
+        }),
+    }),
+    {
+      name: 'task-store', // save store to local storage
+      partialize: (state) => ({ tasks: state.tasks }),
+    },
```
sahil143, line 40: Why do we need to persist this in local storage?

sahil143 (changes requested):

package.json
testcara: The async queue is implemented as a global singleton, using Zustand as a global store aligns perfectly with that architecture. This ensures all components see the same queue state and can update it predictably.

Currently, we just have the scenario for linking secrets to run in queue but thinking that we probably would have other backgrounds jobs for different components in the future, . Zustand’s global store ensures that these jobs are properly queued and managed without conflicts or race conditions.

In comparison, Jotai is powerful for managing component-level state, it doesn’t provide built-in global state persistence. 

For long term consideration,  i choose Zustand. 

BTW, we just installed the Jotai but did not enjoyed it in our projects. Replacing Jotai by Zustand does not bring any regression.

src/components/Secrets/SecretsListView/SecretsListRowTooltip.scss
testcara: To ensure the appearance of the newly added tooltip is consistent with the existing tooltips used throughout our project:

By default, PatternFly tooltips have a black background with white text, as shown below:

Screenshot 2025-05-19 at 18 21 51

You could find more details in the official documentation:: https://v5-archive.patternfly.org/components/tooltip/html/#ws-page-title

However, in our project, the existing tooltips have a white background with black text, as shown here:
[
Screenshot 2025-05-19 at 18 23 49
](url)

This custom style is defined via SCSS, for example in: src/components/PipelineRun/PipelineRunDetailsView/visualization/nodes/PipelineRunNode.scss

Besides the above tooltip example, we enjoy the tooptip help icon quite common in our projects which also shows the white background with black font, as shown below:
[
Screenshot 2025-05-19 at 18 38 10
](url)
I checked the code, the content above is provided by popover. And its style is the default style of the popover content.

src/utils/task-store.ts
testcara: By default, React state and Zustand stores are in-memory only. Refreshing the page clears all states. Persisting this to localStorage allows the app to restore the state even after the user refreshes or closes and reopens the browser. 

i just expect users could feel more reliable and smooth with it and not wonder why the Error disappears just by refreshing pages.

Thinking about we would support 'Edit Secret' soon, users could edit the secrets with relinking to fix the errors.

src/components/Secrets/SecretsListView/SecretsListRowTooltip.scss
sahil143: So for a tooltip, a black background with white text should be good. We shouldn't change the default styles, If you need a white background then please use a popover. Pipeline run visualization is a special case. For table components, we use tooltips with black background.

JoaoPedroPP (commented):
LGTM

Nice use case for zustand

src/components/Secrets/SecretsListView/SecretsList.tsx
```diff
@@ -1,21 +1,30 @@
 import * as React from 'react';
+import { useIsOnFeatureFlag } from '~/feature-flags/hooks';
 import { Table } from '../../../shared';
 import { SecretKind } from '../../../types';
 import SecretsListHeader from './SecretsListHeader';
+import SecretsListHeaderWithComponents from './SecretsListHeaderWithComponents';
```
StanislavJochman, line 6: why we have 2 same things in 2 different files?

rrosatti (approved):
LGTM!

src/components/Secrets/SecretsListView/SecretsListRowWithComponents.tsx
```diff
@@ -54,6 +70,37 @@ const SecretsListRowWithComponents: React.FC<React.PropsWithChildren<SecretsList
...
+        {taskStatus === LinkSecretStatus.Failed ? (
+          <Popover
+            triggerAction="hover"
+            aria-label="Hoverable popover"
+            headerContent={
+              <Alert
+                isInline
+                variant={AlertVariant.warning}
+                title="This error is only available in this browser."
+              />
```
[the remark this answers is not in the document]
testcara: I discarded the local-storage.

src/utils/task-store.ts
testcara: Great questions. i removed the local-storage and enjoy the annotation for the Error msg to ensure all users get the same status.

sahil143 (approved):

janaki29 (approved):

PR #236, opened 2025-05-13
KONFLUX-6148: Fix 'cosign download sbom' command failing due to missing image tag

rrosatti:
## Fixes 

Fixes https://issues.redhat.com/browse/KONFLUX-6148

## Description

The `cosign download sbom` command was failing because the component did not yet have the `status.lastPromotedImage` set. Previously, we used the `getLastestImage()` utility, which falls back to `spec.containerImage` when `status.lastPromotedImage` is unavailable.

However, `spec.containerImage` does not include an image tag, and without a tag, the `cosign download sbom` command fails.

This PR fixes the issue by explicitly using `status.lastPromotedImage` in the `ComponentLatestBuild` component, avoiding the fallback and ensuring the command is executed with a properly tagged image.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

1. Example of broken SBOM command (due to missing image tag in containerImage):

sbom-command-without-image-tag

```
cosign download sbom quay.io/redhat-user-workloads-stage/kkanova-tenant/katka-test-cosign-dd573 
```

2. If we remove the fallback, this command would no longer be displayed:

sbom-no-command-display

3. And for components where `status.lastPromotedImage` is correctly set, the SBOM command displays and works:

sbom-correct-command-with-image-tag

```
cosign download sbom quay.io/redhat-user-workloads-stage/kkanova-tenant/testc@sha256:e3d15797e5537846aeb168be35f367e4388b688a87e4ce45f11cab3bf7e1ecc2
```

## How to test or reproduce?

This issue seems to be happening in some edge cases. The scenario in which we found the issue to be reproduced was when creating a new component and not merging the PR created by PAC. 

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/components/Components/ComponentDetails/tabs/ComponentLatestBuild.tsx
+++ b/src/components/Components/ComponentDetails/tabs/ComponentLatestBuild.tsx
@@ -21,7 +21,6 @@ import { Timestamp } from '../../../../shared/components/timestamp/Timestamp';
 import { useNamespace } from '../../../../shared/providers/Namespace/useNamespaceInfo';
 import { ComponentKind } from '../../../../types';
 import { getCommitsFromPLRs } from '../../../../utils/commits-utils';
-import { getLastestImage } from '../../../../utils/component-utils';
 import CommitLabel from '../../../Commits/commit-label/CommitLabel';
 import { useBuildLogViewerModal } from '../../../LogViewer/BuildLogViewer';
 import ScanDescriptionListGroup from '../../../PipelineRun/PipelineRunDetailsView/tabs/ScanDescriptionListGroup';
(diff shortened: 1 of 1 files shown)


src/components/Components/ComponentDetails/tabs/ComponentLatestBuild.tsx
```diff
@@ -45,7 +44,9 @@ const ComponentLatestBuild: React.FC<React.PropsWithChildren<ComponentLatestBuil
   const [taskRuns, taskRunsLoaded] = useTaskRuns(namespace, pipelineRun?.metadata?.name);
   const buildLogsModal = useBuildLogViewerModal(component);
 
-  const containerImage = getLastestImage(component);
+  // Avoid getLastestImage fallback to spec.containerImage, which lacks image tag
+  // and causes 'cosign download sbom' to fail. Use lastPromotedImage explicitly.
+  const containerImage = component?.status?.lastPromotedImage;
```
testcara, line 49: maybe,  let us fix the problem of getLatestImage directly? just removing the fallback from that function?

rrosatti: I thought about it, but the thing is that in certain scenarios having the fallback works just fine. For example:

https://github.com/user-attachments/assets/520ad10c-bf13-4de1-8f3b-cc2c9778ed42

in those examples it's using the `spec.containerImage` (fallback) and it works fine, because it simply open the quay.io repo 🙂

the issue with the SBOM is that the command fails without the image tag, so it would fail every time it uses the `spec.containerImage` fallback

StanislavJochman (approved):
LGTM

marcin-michal (approved):
LGTM

JoaoPedroPP (approved):
LGTM

The application is working as expected.

PR #238, opened 2025-05-15
fix: handle case where runAfterTasks is undefined

ralphbean:
## Description

I heard from a user that they get stacktrace when trying to view pipelinerun details of an ITS pipeline they kicked off on the cluster:

image

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

## How to test or reproduce?

I'm not sure. I wasn't able to reproduce the failure yet. Perhaps this should be dropped and a test-driven development approach should be taken. Can you advise where in the tests we could try to reproduce the failure before speculating on a fix?

To be clear - I am only *guessing* that the patch here fixes the issue seen by the user in the stacktrace listed above. This may be off the mark.


abhinandan13jan:
/lgtm
Please rebase

ralphbean:
@StanislavJochman @abhinandan13jan I just rebased on `main`.

PR #259, opened 2025-06-02
fix(KFLUXUI-515): show correct error message when accessing namespace without access

marcin-michal:
## Fixes 
[KFLUXUI-515](https://issues.redhat.com/browse/KFLUXUI-515)

## Description
The `NoAccessState` page would display user's own namespace when they tried to access a different namespace, where they don't have access.

The error message now correctly show the name of the namespace which the user is trying to access.

## Type of change

- [X] Bugfix

## How to test or reproduce?
Try to access any namespace, which you do not have access to.

## Browser conformance: 

- [ ] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge


rakshett (approved):
LGTM

PR #268, opened 2025-06-09
fix: make namespace name hyperlink on namespace list page

brianwcook:
Makes NS name a hyperlink so that clicking it will take you to the namespace.


brianwcook:
yran test is working fine locally, not sure why the unit tests are failing in CI ?

src/components/NamespaceList/__tests__/NamespaceListRow.spec.tsx
```diff
@@ -64,6 +64,19 @@ describe('NamespaceListRow', () => {
 
     expect(screen.getByText('Go to the namespace')).toBeInTheDocument();
   });
+
+  it('should make the namespace name a clickable link to the same target as the button', () => {
+    (useApplications as jest.Mock).mockReturnValue([[], true]);
+
+    routerRenderer(<NamespaceListRow columns={[]} obj={mockNamespace} />);
+
+    // Debug: log the DOM output
```
testcara, line 73: for the debug, it is quite useful during the development. 
but we do not recommend to leave the debug before review & merge.
so how about remove the debug?

testcara (approved):
Nice job! LGTM. UI works as expected. just tiny suggestion to remove debug info. Thank you for working on this.:-).

src/components/NamespaceList/NamespaceListRow.tsx
```diff
@@ -48,7 +48,12 @@ const NamespaceListRow: React.FC<React.PropsWithChildren<RowFunctionArgs<Namespa
   return (
     <>
       <TableData className={namespaceTableColumnClasses.name} data-test="app-row-test-id">
-        {obj.metadata.name}
+        <Link
+          to={APPLICATION_LIST_PATH.createPath({ workspaceName: obj.metadata.name })}
+          title="Go to this namespace"
+        >
+          {obj.metadata.name}
+        </Link>
```
sahil143, line 56: We already have a button on the namespace list page, `Go to this namespace`. Do we want to remove that?
In the UI, Any hyperlink on the name takes the user to the details page of that resource and since we dont have a details page for the namespace Ux suggested the addition of the button as an action that takes user to the application list page.

.eslintrc.cjs
```diff
@@ -162,6 +162,7 @@ module.exports = {
       node: {
         extensions: ['.js', '.jsx', '.ts', '.tsx'],
       },
+      typescript: {},
```
sahil143, line 165: Why did you add this?

sahil143 (changes requested):
Thanks for the patch.

testcara:
Just adding hyperlink is a tiny UX change and if users consider it is more convenient, it looks good to me.
About replacing the button with hyperlink, i guess we need UX's double-confirm.
Personally, i like the hyperlink way which matches my habit, just clicking what i see. :-).
Forwarding the discussion to ui  slack channel now.. :-).

src/components/NamespaceList/NamespaceListRow.tsx
pankajshivpuje: @brianwcook This design was a bit tricky. Linking a namespace given an impressions that it will take user to the details view of the namespace but that's not the case. The goal of the button was to let user set the context by selecting an action on the namespace. My original design had that action on the kebab menu but the feedback I received was that it would a bit hidden and it would helpful to have it more visible in the table. The result is the current design that you are looking at. I agree, it looks a bit clunky but my understanding was that it was a stopgap solution until we are able to support the creation of namespace within Konflux.

pankajshivpuje:
The easiest approach would be to fix the actions column. We can make the namespace name as a link when we have a detailed namespace page available. This will be consistent with approach we have taken for application, secrets, releases, components etc in Konflux. Please see the attached screenshots for reference.
Namespace list view#5
Namespace list view#4

brianwcook:
Why would I want to look down the left column, then scan all the way across the page and then click on some dots, and then finally click on a 'go to namespace' entry in a list instead of just clicking on the namespace name that I want to navigate to?

That seems even worse than what is there now. It is three clicks instead of 2, mine is 1 click.

cit1zen:
Looking at this debate, we should probably change the way how it is handled everywhere else as I would expect that I can click on the name to get to the object/place from the list. We do it for applications, why not everywhere else. I will add it to our Monday debate.

cit1zen:
Reading this MR and the debate, I believe that we should make names into links that switch you to the namespace similarly to the OpenShift.

In the old model the link will move you to the page that lists application. In the new model it can move you to overview, where we will list the most critical issues from the issues dashboard and maybe the latest activity or something.

tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
.../page-routes/pipeline.tsx","./src/routes/page-routes/release.tsx","./src/models/release-plan-admission.ts","./src/components/releaseservice/releaseservice.tsx","./src/components/pageaccess/pageaccesscheck.tsx","./src/components/pageaccess/withpageaccesscheck.tsx","./src/components/releaseservice/releaseserviceemptystate.tsx","./src/components/releaseservice/releaseplan/releaseplanlistheader.tsx","./src/components/releaseservice/releaseplan/releaseplan-actions.tsx","./src/components/releaseservice/releaseplan/releaseplanlistrow.tsx","./src/components/releaseservice/releaseplan/releaseplanlistview.tsx","./src/types/release-plan-admission.ts","./src/hooks/usereleaseplanadmissions.ts","./src/components/releaseservice/releaseplanadmission/releaseplanadmissionlistheader.tsx","./src/components/releaseservice/releaseplanadmission/releaseplanadmission-actions.tsx","./src/components/releaseservice/releaseplanadmission/releaseplanadmissionlistrow.tsx","./src/components/releaseservice/releaseplanadmission/releaseplanadmissionlistview.tsx","./src/components/releaseservice/releaseplan/triggerrelease/form-utils.ts","./node_modules/@types/lodash/compact.d.ts","./node_modules/@types/lodash/values.d
```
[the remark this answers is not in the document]
brianwcook: done!

testcara (approved):
LGTM  UI works as expected. Thank you for the discussion and work.

rrosatti (approved):
LGTM. UI is working fine :)

brianwcook:
What is needed to get this merged?

pankajshivpuje:
LGTM

PR #318, opened 2025-07-07
fix(KFLUXUI-578): don't allow incomplete component relationships

marcin-michal:
Assisted-by: Cursor

## Fixes 
https://issues.redhat.com/browse/KFLUXUI-578

## Description
It was possible for a user to create an incomplete component relationship, only specifying the source and not the target or wise versa. Due to how the form for adding a new relationship works (there is always at least one, empty relationship), the schema is valid only if the relationships is fully empty or fully defined.

## Type of change

- [ ] Feature
- [X] Bugfix
- [ ] Build related changes

Before:

https://github.com/user-attachments/assets/1947f320-6bfb-4057-b4df-7f289cb01d69

After:

https://github.com/user-attachments/assets/9d1acc46-296d-481f-ae06-90db95a093c2

## How to test or reproduce?
Go to the application page, click the `Actions` button and click `Define component relationships`. Select only the source and leave the target blank.

### DIFF
--- a/src/components/ComponentRelation/__tests__/utils.spec.ts
+++ b/src/components/ComponentRelation/__tests__/utils.spec.ts
@@ -3,6 +3,7 @@ import {
   componentRelationValidationSchema,
   computeNudgeDataChanges,
   transformNudgeData,
+  toCustomBoolean,
 } from '../utils';
 
 describe('computeNudgeDataChanges', () => {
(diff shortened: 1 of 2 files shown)


rrosatti (approved):
LGTM!

marcin-michal:
> LGTM
> 
> Minor question: I was playing with the modal and I found out I am able to add as many empty relations as I want, does it make sense to make the `Add another component relationship` available while we don't have one valid relation?
> image

No, it does not make sense IMO :smile: I think it is a part of a bigger problem with the way how the relationships are created/removed, as I discussed with you yesterday on the standup, and should probably be handled together as a whole.

testcara (approved):
LGTM. UI works as expected.

PR #436, opened 2025-09-25
fix(KONFLUX-10107): assure component list view is using the lastest build

JoaoPedroPP:
Assisted-by: Cursor

## Fixes 

https://issues.redhat.com/browse/KONFLUX-10107

## Description

Query the pipelines only for the required components and sort the output to assure `useLatestBuildPipelines()` return the lastest pipelinerun an specific component.

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes

## How to test or reproduce?

1. Component latest build info in Components tab should be the same in Activity pipelines tab

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


StanislavJochman (dismissed):
LGTM UI works as expected

StanislavJochman (approved):

rrosatti (approved):
LGTM! nice job :)

marcin-michal (approved):
LGTM

src/hooks/useLatestBuildPipelines.ts
```diff
@@ -32,9 +32,14 @@ export const useLatestBuildPipelines = (
             [PipelineRunLabel.APPLICATION]: applicationName,
             [PipelineRunLabel.PIPELINE_TYPE]: PipelineRunType.BUILD,
           },
+          ...((componentNames?.length ?? 0) > 0 && {
+            matchExpressions: [
+              { key: PipelineRunLabel.COMPONENT, operator: 'In', values: componentNames },
+            ],
+          }),
```
sahil143, line 39: is this filter parsed properly and added to the tekton results API URL? Also, we need to inform the pipelines team about the changed in filter to prevent slow api response time from tekton results.

JoaoPedroPP: You are right, just checked in the requests and it wasn't there. In fact we do not need this part since we are going to use find method anyway.

We can also update this to use `usePipelineRunsV2` which should be able use matchExpression in the call. If not

janaki29 (approved):

marcin-michal (approved):

PR #547, opened 2025-11-18
fix: improve performance of kubearchive queries

maruiz93:
Add creationTimestampAfter and creationTimestampBefore when querying KubeArchive based on the related
creationTimestamp information.

Created by: Cursor AI

## Fixes 

https://issues.redhat.com/browse/KONFLUX-11074

## Description

In some situations where there are too many resources in a namespace compared to others, the KubeArchive database choose not to use some indexes meaning that the performance of the queries decreases at a point of not being able to load the page.
Adding more filters improves the performance so we added the `creationTimestampAfter` and `creationTimestampBefore` index.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

## How to test or reproduce?

The bug is not easy no reproduce, it needs a specific data distribution in the kubearchive cluster to make the engine not be performant and choosing to use a sequencial search.

The fix can be tested by checking that the query now includes `creationTimestampBefore` and `creationTimestampAfter` fields to the kubearchive endpoint and that the queries with Tekton Results keep working

image

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

I didn't try different browsers but this change doesn't affect the interface.

I am not familiar with the codebase or with React, so I didn't do the changes and I totally relied on the AI. Please, review the changes thoroughly to ensure things are done as it should.
Feel free to close and reopen the PR if the changes aren't compliant with the project expectations.

## Release Notes

* **New Features**
  * Added loading indicators to task runs and security tabs for better visibility during data retrieval.

* **Performance**
  * Optimized data queries using pipeline run creation timestamps to improve filtering efficiency and reduce unnecessary data fetching.

* **Bug Fixes**
  * Enhanced accuracy of task run and scan result filtering by incorporating creation timestamp context.

### DIFF
--- a/src/hooks/useScanResults.ts
+++ b/src/hooks/useScanResults.ts
@@ -91,12 +91,16 @@ const processScanTaskRuns = (taskRuns: TaskRunKind[]): ScanResults => {
   return resultObj;
 };
 
-export const useScanResults = (pipelineRunName: string): [ScanResults, boolean, unknown] => {
+export const useScanResults = (
+  pipelineRunName: string,
+  pipelineRunCreationTimestamp?: string,
+): [ScanResults, boolean, unknown] => {
   const namespace = useNamespace();
   const [taskRuns, loaded, error, getNextPage, nextPageProps] = useTaskRunsForPipelineRuns(
     namespace,
     pipelineRunName,
-    undefined,
+    undefined, // taskName
+    pipelineRunCreationTimestamp,
     false, // Don't watch - scan results are for completed pipeline runs
   );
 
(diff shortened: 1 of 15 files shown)


src/hooks/useScanResults.ts
```diff
@@ -134,9 +138,34 @@ const dataSelectorForScanResults = (data: InfiniteData<TaskRunKind[], unknown>):
...
+      );
+      return undefined;
+    }
+
+    try {
+      const creationTime = new Date(pipelineRunCreationTimestamp);
+      const timestampAfter = pipelineRunCreationTimestamp;
+      const timestampBefore = new Date(creationTime.getTime() + 24 * 60 * 60 * 1000).toISOString();
+
+      return `creationTimestampAfter=${timestampAfter},creationTimestampBefore=${timestampBefore}`;
```
rrosatti: hmm, maybe we could reuse the `convertFilterToKubearchiveSelectors` utility? 🤔 wdyt @sahil143?

maruiz93:
I am closing the PR as we found a solution solve the related issue.

This is still consider a potential improvement, but we will create a separate ticket and explore where and if it makes sense to add extra filters

PR #619, opened 2025-12-17
KFLUXUI-938: prevent archived resources from disappearing without page refresh

rrosatti:
## Fixes 

Fixes https://issues.redhat.com/browse/KFLUXUI-938

## Description

Fixes issue where archived PipelineRuns and TaskRuns disappear from the UI when resources transition from cluster to archive, requiring a page refresh to see archived data.

Main Changes:

- **preserve removed TaskRuns**: add `etcdRunsRef` logic to preserve TaskRuns removed from cluster (archiving scenario), ensuring they remain available without page refresh.

- **archive querying when cluster returns 404 errors**: enable archive service queries only when cluster returns 404 error for single resource fetches. Cluster data takes priority when available.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

I'm not providing any screen-recording here because it takes a couple of minutes from triggering a new PLR to having it archived.

## How to test or reproduce?

There are a couple of places to be checked:

1. PLR list
2. PLR details
3. PLR - TaskRuns list
4. TaskRun details

In all above places, you need to trigger a new PLR and wait for when it will be archived. When it happens, the page should keeps showing the correct info. Prior to this PR's changes, the page was either showing a 404 error (TaskRun Details / PipelineRun Details) OR in case of the lists (PipelineRuns list and TaskRuns list), the resource was disappearing.

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## Release Notes

* **New Features**
  * Added intelligent fallback handling for pipeline and task run queries when primary data sources are unavailable, automatically switching to alternative data sources.

* **Tests**
  * Expanded test coverage for error scenarios and fallback data source behavior, including refetch logic and data source prioritization.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/hooks/useTaskRunsV2.ts
+++ b/src/hooks/useTaskRunsV2.ts
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { InfiniteData } from '@tanstack/react-query';
-import { uniqBy } from 'lodash-es';
+import { differenceBy, uniqBy } from 'lodash-es';
 import { useIsOnFeatureFlag } from '~/feature-flags/hooks';
 import { TQueryInfiniteOptions } from '~/k8s/query/type';
 import { WatchK8sResource } from '~/types/k8s';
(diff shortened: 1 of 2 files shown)


src/hooks/usePipelineRunsV2.ts
```diff
@@ -241,6 +240,26 @@ export const usePipelineRunsV2 = (
...
+    if (clusterBecameEmpty) {
+      if (kubearchiveEnabled) {
+        void kubearchiveResult.refetch();
+      } else {
+        void trRefetch();
+      }
+    }
+
+    prevClusterLengthRef.current = currentLength;
+  }, [processedClusterData?.length, kubearchiveEnabled, kubearchiveResult, trRefetch]);
```
sahil143: I dont think we need to refetch here because `etchRunsRef` keeps the resources that were deleted from cluster. We need to make sure that the resource is present in `etcdRunsRef`.

sahil143 (changes requested):

src/hooks/usePipelineRunsV2.ts
rrosatti: That's true, I didn't pay attention to the `etcdRunsRef` logic. I figured it out the issue and fixed it. Basically, inside the `runs` memo, we had this logic at the beginning:

```tsx
    if (!etcdRuns.length) {
      return etcdRuns;
    }
```

but it should be:

```tsx
    if (!etcdRuns) {
      return etcdRuns;
    }
```

src/hooks/useTaskRunsV2.ts
```diff
@@ -101,6 +107,26 @@ export const useTaskRunsV2 = (
...
+    if (clusterBecameEmpty) {
+      if (enableKubearchive) {
+        void kubearchiveQuery.refetch();
+      } else {
+        void tektonRefetch();
+      }
+    }
+
+    prevClusterLengthRef.current = currentLength;
+  }, [processedClusterData?.length, enableKubearchive, kubearchiveQuery, tektonRefetch]);
```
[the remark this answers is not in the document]
rrosatti: I followed the same approach as we have for `usePipelineRuns` hook, so we don't need this refetch logic. PR will be updated shortly

sahil143 (approved):
just one comment, Code looks good. Need UI testing.

PR #711, opened 2026-02-17
[KFLUXUI-1070] refactor(e2e-tests/utils/Applications): use expect() in checkPipelineStatuses

rrosatti:
## Fixes 

Fixes https://issues.redhat.com/browse/KFLUXUI-1070

## Description

In this PR we're fixing the logic in the `checkPipelineStatuses` function, so it uses Cypress `expect()` callback instead of returning `true/false`. It seems Cypress would not fail/throw if it returns false.

Also, we are adding a logic  to check if the -on-pull pipeline run is visible before merging the onboarding PR. It's a proposed approach to prevent https://issues.redhat.com/browse/KFLUXUI-796 from happening. 

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Before (notice on-pull PLR status is `Running`, and the test to check if the status is Cancelled succeeded):

e2e-cancelling-or-cancelled-check-ISSUE

After (notice on-pull PLR status is `Running`, and the test to check if the status is Cancelled correctly failed):

e2e-cancelling-or-cancelled-check-FIX

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## Summary by CodeRabbit

* **Tests**
  * Added a verification step to confirm pipeline runs are visible from a component’s details view.
  * Refined pipeline status checks for more reliable waits and clearer failure messages.
  * Added page-level helpers to simplify locating and asserting pipeline run entries during end-to-end tests.

### DIFF
--- a/e2e-tests/support/pages/ComponentDetailsPage.ts
+++ b/e2e-tests/support/pages/ComponentDetailsPage.ts
@@ -15,15 +15,16 @@ export class ComponentDetailsPage {
   }
 
   static verifyPipelineRunIsVisible(plrName: string) {
-    cy.get(activityTabPO.clickTab, { timeout: 30000 }).click();
-    Common.waitForLoad();
-    cy.get(compActivityPipelinerunsTabPO.clickTab, { timeout: 30000 }).click();
+    cy.get(activityTabPO.clickTab).click();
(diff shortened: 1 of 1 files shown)


marcin-michal (approved):
LGTM

PR #768, opened 2026-03-10
fix(#765): createRepoUrl() returns null for GitLab and Forgejo providers

harshakumar25:
## Fixes 
Fixes https://github.com/konflux-ci/konflux-ui/issues/765

## Description
Found an issue where branch and PR links weren't working for GitLab and Forgejo commits. The problem was in `createRepoUrl()` - it was checking the provider type first and returning null for anything that wasn't GitHub, even when the commit already had a valid `repoURL`.

**What changed:**
- Fixed `createRepoUrl()` to check if `repoURL` exists first before looking at provider type
- Added support for building `gitlab.com` URLs when only org/repo info is available
- Updated `createRepoBranchURL()` to handle different URL patterns:
  - GitLab: `/-/tree/{branch}`
  - Forgejo: `/src/branch/{branch}`
  - GitHub: `/tree/{branch}` (unchanged)
- Updated `createRepoPullRequestURL()` for provider-specific formats:
  - GitLab: `/-/merge_requests/{number}` (they call them merge requests)
  - Forgejo: `/pulls/{number}` 
  - GitHub: `/pull/{number}` (unchanged)
- Added test coverage for both new providers

**Self-hosted note:** Forgejo is always self-hosted, so we can't build URLs without the `repoURL` field. The backend populates this from annotations, so it should work fine.

Screenshot 2026-03-10 at 11 54 08 PM

Screenshot 2026-03-10 at 11 54 45 PM

## How to test or reproduce?

Run tests:
```bash
yarn test commits-utils.spec.ts

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for multi-provider URL construction across GitHub, GitLab, and Forgejo.

* **Refactor**
  * Provider-aware URL generation for repositories, branches, commits, and pull requests with improved fallback logic.

* **Bug Fixes**
  * Commit links are now conditional: when no commit URL is available, the commit label and actions render as non-clickable items.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

**Review by marcin-michal — CHANGES_REQUESTED** [review]
Thanks for the contribution! The bug you pointed out is valid.

However, we recently updated `src/utils/git-utils.tsx` which already has a data-driven provider config covering the same URL patterns (branch paths, commit paths) for GitHub, GitLab, Forgejo, Gitea, Codeberg, and Bitbucket (including self-hosted instance detection via hostname parsing).

Rather than duplicating provider knowledge with if/else chains in commits-utils.ts, I think it would be preferable to:
- Add `pullRequestPath` to the existing provider config in git-utils.tsx
- Refactor createRepoBranchURL / createRepoPullRequestURL in commits-utils.ts to delegate to git-utils.tsx
- Also fix the shaURL fallback (lines 53, 276) which still hardcodes `/commit/` (broken for GitLab)

This keeps a single source of truth for URL patterns and gets us Bitbucket and self-hosted GitLab/Gitea support.

**Review by marcin-michal — COMMENTED** [review]
the code seems pretty good to me, fix the code rabbit suggestion and unsafe argument of type any assignment

**Comment by harshakumar25** [comment]
hi @rrosatti ,
can you please review this pr.
Thanks !

**Review by StanislavJochman — APPROVED** [review]
LGTM

**Comment by harshakumar25** [comment]
hi @marcin-michal ,
i think everything is looking best (title and description ) if need any modification please let me know else i will happy to get approved/merge on this 
Thanks !

File: src/utils/commits-utils.ts [hunk 1]
```diff
@@ -48,16 +54,16 @@ export const createCommitObjectFromPLR = (plr: PipelineRunKind): Commit => {
...
-  const shaTitle =
-    plr.metadata.annotations?.[PipelineRunLabel.COMMIT_SHA_TITLE_ANNOTATION] || 'manual build';
   const gitProvider =
     plr.metadata.labels?.[PipelineRunLabel.COMMIT_PROVIDER_LABEL] ||
     plr.metadata.annotations?.[PipelineRunLabel.COMMIT_PROVIDER_LABEL] ||
     plr.metadata.labels?.[PipelineRunLabel.TEST_COMMIT_PROVIDER_LABEL] ||
     plr.metadata.annotations?.[PipelineRunLabel.TEST_COMMIT_PROVIDER_LABEL];
+  const shaURL =
+    plr.metadata.annotations?.[PipelineRunLabel.COMMIT_URL_ANNOTATION] ||
+    createGitCommitURL(repoURL, commitSHA, gitProvider);
```
marcin-michal, line 64: both of these could be `null`, so the `shaURL` type should be updated to allow it in `src/types/commit.ts` and all its downstream consumers

src/utils/git-utils.tsx
```diff
@@ -6,72 +6,159 @@ import { GitlabIcon } from '@patternfly/react-icons/dist/esm/icons/gitlab-icon';
 import gitUrlParse from 'git-url-parse';
 import ForgejoLogo from '../shared/assets/forgejo-logo.svg';
 
-type ProviderConfig = {
+type GitProviderConfig = {
```
marcin-michal, line 9: there's a lot of changes in `git-utils.tsx` but no changes in the spec file for it. update the unit tests to cover all changes

marcin-michal (changes requested):
@harshakumar25

PR #784, opened 2026-03-23
KFLUXUI-1153 - Fix: "PLR Graph displays task failures during archival transition"

rrosatti:
## Fixes 

Fixes https://redhat.atlassian.net/browse/KFLUXUI-1153

## Description

In this PR we're fixing the PLR deails visualization showing every task as Failed (red) when a run moves from "in-cluster / etcd" to archived (Tekton Results / KubeArchive) while the page stays open.

After some investigation I noticed the following problem: 

- after the PLR is fetched/loaded from the archive, taskRuns could stay empty
- the `PipelineRunVisualization` component builds task status from TaskRuns; when the list is empty, appendStatus falls back to the overall PipelineRun status for every task, so a failed PLR makes all tasks look failed.
- root cause: `useTaskRunsForPipelineRuns` passed `staleTime: Infinity` into `useTaskRunsV2` for the Tekton Results / KubeArchive list queries
  - that empty result was cached as never stale, so when `TaskRuns` disappeared from the cluster watch, we did not refetch the archive side and the merged list stayed empty

~~So, I tried the following fix: removed `staleTime: Infinity` so archive `TaskRun` lists can become stale and refetch, allowing `TaskRuns` to populate after archive transition~~

updated fix: replace static `staleTime: Infinity` with a function -> non-empty results are cached forever (task runs don't change once completed), empty results expire immediately so they get refetched after the archive transition

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Before (kubearchive call happens at 00:39):

https://github.com/user-attachments/assets/ff0e24e8-cbfa-4fcd-af19-4a4c4b084c08

After (kubearchive call happens at 00:39):

https://github.com/user-attachments/assets/12ac015d-9408-4715-8223-9c0b6ea86775

## How to test or reproduce?

- trigger a new PipelineRun (preferably try to make one of its TaskRuns to fail)
- stay in the PLR details page
- wait until the archival happens
- with this PR, the TaskRuns will be refetched and the `PipelineRunVisualization` will show the correct info
- :coffee:

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## Summary by CodeRabbit

* **Refactor**
  * Improved caching behavior for task-run queries: cache lifetime is now conditional based on whether previously fetched data exists, reducing stale data and improving responsiveness when navigating between pipelines.

### DIFF
--- a/src/hooks/useTaskRunsV2.ts
+++ b/src/hooks/useTaskRunsV2.ts
@@ -248,7 +248,10 @@ export const useTaskRunsForPipelineRuns = (
       selector,
       watch,
     },
-    { staleTime: Infinity, enabled: !!(namespace && pipelineRunName) },
+    {
+      enabled: !!(namespace && pipelineRunName),
+      staleTime: (query) => (query?.state?.data?.pages?.flat()?.length ? Infinity : 0),
+    },
   );
 
   const sortedTaskRuns = React.useMemo(() => sortTaskRunsByTime(taskRuns), [taskRuns]);


StanislavJochman (dismissed):
LGTM works as expected

testcara (approved):
LGTM, tested it, it works well locally, I could see it is refreshing automatically.

src/hooks/useTaskRunsV2.ts
```diff
@@ -248,7 +248,7 @@ export const useTaskRunsForPipelineRuns = (
       selector,
       watch,
     },
-    { staleTime: Infinity, enabled: !!(namespace && pipelineRunName) },
+    { enabled: !!(namespace && pipelineRunName) },
```
sahil143: This fixes the immediate issue but introduces a new one. On the pipelineRuns list page, task runs will no longer be archived. Instead, they’ll be refetched every time the stale window expires. Since this is an infinite-scrolling list where rows stay mounted, API calls scale with both the number of loaded rows and how long the user stays on the page. For example, 100 loaded rows with a 30s stale time over a 5-minute session results in roughly 600 API calls instead of the current 100. The longer the session, the worse it gets

sahil143 (changes requested):

milantaky (approved):
LGTM, working as expected

src/hooks/useTaskRunsV2.ts
```diff
@@ -248,7 +248,10 @@ export const useTaskRunsForPipelineRuns = (
       selector,
       watch,
     },
-    { staleTime: Infinity, enabled: !!(namespace && pipelineRunName) },
+    {
+      enabled: !!(namespace && pipelineRunName),
+      staleTime: (query) => (query?.state?.data?.pages?.flat()?.length ? Infinity : 0),
```
sahil143, line 253: you can not keep stale time 0, that would mean data is always stale even when it is just fetched. it should be undefined.

rrosatti: maybe I misunderstood it, but from the docs I thought that if we set to `undefined` it'd fallback to `0` anyways 🤔 

from the `useQuery` docs:
```
staleTime: number | Infinity
Optional
Defaults to 0
The time in milliseconds after data is considered stale. This value only applies to the hook it is defined on.
If set to Infinity, the data will never be considered stale
```

ref: https://tanstack.com/query/v4/docs/framework/react/reference/useQuery

StanislavJochman (approved):

sahil143 (approved):

PR #816, opened 2026-04-09
fix(KFLUXUI-1156): show loading indicator while logs are being processed

rrosatti:
## Fixes 

Fixes https://redhat.atlassian.net/browse/KFLUXUI-1156

## Description

In this PR we're updating the `Logs` component to track the processing logs state and pass it to `LogViewer`'s `isLoading` prop. 

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Before:

https://github.com/user-attachments/assets/59623099-614d-4488-b761-a6d7746a298e

After:

https://github.com/user-attachments/assets/807f4906-a505-49d0-b458-6192b216ddf9

## How to test or reproduce?

You can simply check logs from any PLRs, you'll notice that there'll be a loading spinner at the top while the logs are being loaded.

If you want to try to simulate "loading a huge log", you can add the following to `src/shared/components/pipeline-run-logs/logs/Logs.tsx`:

```tsx
  React.useEffect(() => {
    fetch('/mock-huge-log.log')
      .then((res) => res.text())
      .then((logData) => {
        // eslint-disable-next-line no-console
        console.log(
          `[MOCK] Loaded real log file: ${logData.length} chars, ~${logData.split('\n').length} lines`,
        );
        // Inject into only the first container (simulates one step with huge logs)
        if (containers.length > 0) {
          appendLog(containers[0].name, logData);
        }
      })
      // eslint-disable-next-line no-console
      .catch((err) => console.error('[MOCK] Failed to load mock log file:', err));
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

You just need to create a file called `mock-huge-log.log` and put it inside the `public/` folder. This file will contain the huge log, so you can somehow foce it to be loaded in the LogViewer :)

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## Summary by CodeRabbit

* **Bug Fixes**
  * Loading indicator now reflects both external loading and internal log-processing, preventing stale or missing spinners during log formatting.

* **Style**
  * Adjusted spacing for the log viewer's loading display for improved visual alignment.

* **Tests**
  * Added tests verifying loading state during debounced log processing and when no log sources are present.

### DIFF
--- a/src/shared/components/pipeline-run-logs/logs/Logs.tsx
+++ b/src/shared/components/pipeline-run-logs/logs/Logs.tsx
@@ -3,6 +3,7 @@ import { useTranslation } from 'react-i18next';
 import { Base64 } from 'js-base64';
 import { useIsOnFeatureFlag } from '~/feature-flags/hooks';
 import { KUBEARCHIVE_PATH_PREFIX } from '~/kubearchive/const';
+import { useDebounceCallback } from '~/shared/hooks/useDebounceCallback';
 import { ResourceSource } from '~/types/k8s';
 import { commonFetchText } from '../../../../k8s';
 import { getK8sResourceURL, getWebsocketSubProtocolAndPathPrefix } from '../../../../k8s/k8s-utils';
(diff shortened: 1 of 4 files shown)


src/shared/components/pipeline-run-logs/logs/Logs.tsx
```diff
@@ -186,10 +187,28 @@ const Logs: React.FC<LogsProps> = ({
...
-    () => processLogs(logSources, containers),
-    [logSources, containers],
-  );
+  const [formattedLogs, setFormattedLogs] = React.useState('');
+  const [processingLogs, setProcessingLogs] = React.useState(false);
+  const containersRef = React.useRef(containers);
+  containersRef.current = containers;
+
+  const processAndSetLogs = useDebounceCallback(() => {
+    const result = processLogs(logSources, containersRef.current);
```
sahil143, line 190: All processLog does is split the string and then join them again for each container. Not sure how this processAndSetLogs is useful. Main problem comes from processLogs.

sahil143 (changes requested):

src/shared/components/pipeline-run-logs/logs/Logs.tsx
rrosatti: I believe keeping the debounce is still worth it. even with `processLogs` optimized, it still scans the full log string on every call. Without the debounce, that would happen on every incoming ws message.

One improvement I just did was to replace the "split/map/join" logic with a single regex (in the `processLogs` func).

StanislavJochman (approved):
LGTM UI works as expected

milantaky (approved):
LGTM, works as expected

sahil143 (approved):

PR #817, opened 2026-04-09
feat(KFLUXUI-1197): create text filtering util

marcin-michal:
Assisted-by: Cursor

## Fixes 
[KFLUXUI-1197](https://redhat.atlassian.net/browse/KFLUXUI-1197)

## Description
Utility for consistent text filtering, with the ability to:
- trim whitespaces,
- case (in)sensitive
- subsequent search (`plr` -> `pipelinerun`) 

The current filtering will be replaced with this utility in a different PR for more clarity

## Type of change

- [X] Feature
- [ ] Bugfix
- [ ] Build related changes

[KFLUXUI-1197]: https://redhat.atlassian.net/browse/KFLUXUI-1197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Configurable text matching and list filtering: case-sensitive option, optional trimming of queries, and fuzzy (ordered-subsequence) matching. Filters return the original list when the query is empty and handle non-string values, Unicode, and literal special characters.

* **Tests**
  * Added comprehensive tests for defaults, option combinations, robustness with null/undefined/non-string inputs, empty queries/lists, and Unicode/special-character scenarios.

### DIFF
--- a/src/utils/text-filter-utils.ts
+++ b/src/utils/text-filter-utils.ts
@@ -1,0 +1,124 @@
+export type TextFilterOptions = {
+  caseSensitive?: boolean;
+  trim?: boolean;
+  /** Use subsequence (fuzzy) matching: all characters of the search term must
+   *  appear in order within the value, but not necessarily contiguously.
+   *  E.g. "cbp" matches "component-build-pipeline". Defaults to `false`. */
+  fuzzy?: boolean;
+};
+
+type Searchable = string | number | boolean | null | undefined;
+
+/**
+ * Safely coerces a primitive value to a string for text comparison.
+ */
+const toSearchableString = (value: Searchable): string => {
+  if (value === null) {
+    return '';
+  }
+
+  if (value === undefined) {
+    return '';
+  }
+
+  if (typeof value === 'string') {
+    return value;
+  }
+
+  return String(value);
+};
+
+/**
+ * Subsequence match: every character in `search` appears in `text` in order,
+ * but not necessarily contiguously. Both strings must already be in the
+ * desired case before calling this.
+ */
+const subsequenceMatch = (text: string, search: string): boolean => {
+  let si = 0;
+
+  for (let ti = 0; ti < text.length && si < search.length; ti++) {
+    if (text[ti] === search[si]) {
+      si++;
+    }
+  }
+
+  return si === search.length;
+};
+
+/**
+ * Checks whether a **value** matches a **search term** based on the provided options.
+ *
+ * Key behaviours:
+ * - Non-string values (numbers, booleans, `null`, `undefined`) are coerced to
+ *   a string before comparison, so calling code never has to worry about runtime
+ *   errors from e.g. `(42).toLowerCase()`.
+ * - An empty (or whitespace-only, when `trim` is enabled) search term matches
+ *   everything.
+ * - When `fuzzy` is true, uses subsequence matching: all characters of the
+ *   search term must appear in order within the value.
+ *
+ * @param value      - The value to search in.
+ * @param searchTerm - The term to look for.
+ * @param options    - Optional search configuration.
+ * @returns `true` when the value matches the search term.
+ *
+ * @example
+ * ```ts
+ * textMatch('Hello World', 'hello');                        // true  (case-insensitive)
+ * textMatch('Hello World', 'hello', { caseSensitive: true }); // false
+ * textMatch(42, '4');                                       // true  (number → "42")
+ * textMatch('Hello World', 'hwd', { fuzzy: true });         // true  (subsequence)
+ * textMatch(null, 'x');                                     // false
+ * textMatch('anything', '  ');                              // true  (empty after trim)
+ * ```
+ */
+export const textMatch = (
+  value: Searchable,
+  searchTerm: string,
+  options: TextFilterOptions = {},
+): boolean => {
+  const { caseSensitive = false, trim = true, fuzzy = false } = options;
+
+  let text = toSearchableString(value);
+
+  if (trim) {
+    searchTerm = searchTerm.trim();
+  }
+
+  if (!searchTerm) {
+    return true;
+  }
+
+  if (!text) {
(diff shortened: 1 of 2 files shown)


rrosatti (dismissed):
LGTM

rrosatti (dismissed):

src/utils/text-filter-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+   *  E.g. "cbp" matches "component-build-pipeline". Defaults to `false`. */
+  fuzzy?: boolean;
+};
+
+/**
+ * Safely coerces a value to a string for text comparison.
+ * Handles strings, numbers, nullish values, and other types.
+ */
+const toSearchableString = (value: unknown): string => {
+  if (value == null) {
```
sahil143:
```suggestion
  if (value === null) {
```

src/utils/text-filter-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+   *  appear in order within the value, but not necessarily contiguously.
+   *  E.g. "cbp" matches "component-build-pipeline". Defaults to `false`. */
+  fuzzy?: boolean;
+};
+
+/**
+ * Safely coerces a value to a string for text comparison.
+ * Handles strings, numbers, nullish values, and other types.
+ */
+const toSearchableString = (value: unknown): string => {
```
sahil143:
```suggestion
const toSearchableString = (value: string): string => {
```
we expect value to be a string, this util is to avoid runtime errors

src/utils/text-filter-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+ * textMatch('Hello World', 'hello', { caseSensitive: true }); // false
+ * textMatch(42, '4');                                       // true  (number → "42")
+ * textMatch('Hello World', 'hwd', { fuzzy: true });         // true  (subsequence)
+ * textMatch(null, 'x');                                     // false
+ * textMatch('anything', '  ');                              // true  (empty after trim)
+ * ```
+ */
+export const textMatch = (
+  value: unknown,
+  searchTerm: unknown,
```
sahil143: should be strings
```suggestion
  value: string,
  searchTerm: string,
```

src/utils/text-filter-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+};
+
+/**
+ * Filters an array of items by matching a search term against a text field
+ * extracted via `accessor`. Returns the **original array reference** when the
+ * search term is empty so that `useMemo` consumers avoid unnecessary rerenders.
+ */
+export const filterByText = <T>(
+  items: T[],
+  searchTerm: unknown,
```
sahil143:
```suggestion
  searchTerm: string,
```

src/utils/text-filter-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+  accessor: (item: T) => unknown,
```
sahil143:
```suggestion
  accessor: (item: T) => string,
```
we are comparing strings, so accessor should return string

sahil143 (changes requested):

src/utils/text-filter-utils.ts
marcin-michal: this is so it is true for both undefined and null variables

src/utils/text-filter-utils.ts
marcin-michal: but that's exactly why we can't expect the value to be a string, right? we don't know whether it's string, number, array, or bool, so we transform it into a string. Otherwise `.toLowerCase()`, `.trim()` would cause runtime errors

src/utils/text-filter-utils.ts
marcin-michal: I don't think this is the case in the current code base, but what if we are searching in an array of numbers? I think this utility should be as general and flexible as possible, and not allow only strings.

src/utils/text-filter-utils.ts
sahil143: we should explicitly check for undefined and null if that's the case. having `==` can have unexpected consequences.

src/utils/text-filter-utils.ts
sahil143: Agreed, we should have checks to prevent runtime error but typescript is for static checks, it's basically what you expect as the value. if it can be number then we should use `string | number' marking it unknown gives the idea that we can pass objects and arrays too

src/utils/text-filter-utils.ts
sahil143: The function is called ‎`filterByText`, which implies string comparison. The search term from a text input will always be a string, so the accessor should also return a string.

If the accessor returns a number, calling ‎`toLowerCase()` on it would cause a runtime error. To properly support both types, you’d need separate comparison logic, ‎`===` for numbers and ‎`includes` for strings. which adds complexity this utility shouldn’t own.

Typing the accessor as ‎`(item: T) => string` makes the contract clear: consumers are responsible for converting their values to strings before passing them in else accessor can also return arrays and objects. If there’s a need to filter by numeric equality, that should be a separate utility.

src/utils/text-filter-utils.ts
marcin-michal: updated

src/utils/text-filter-utils.ts
marcin-michal: updated to more narrow type signature

src/utils/text-filter-utils.ts
marcin-michal: updated to more narrow type signature

sahil143 (approved):

StanislavJochman (approved):
LGTM

rrosatti (approved):
LGTM

PR #829, opened 2026-04-22
ci(KFLUXUI-1223): switch to operator Konflux-ci deployment

marcin-michal:
Assisted-by: Cursor

## Fixes 
[KFLUXUI-1223](https://redhat.atlassian.net/browse/KFLUXUI-1223)

## Description
Replaces the legacy konflux-ci deployment with the new operator-based deployment

## Type of change

- [ ] Feature
- [ ] Bugfix
- [X] Build related changes

[KFLUXUI-1223]: https://redhat.atlassian.net/browse/KFLUXUI-1223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Simplified CI/workflow and local deployment to a single local deploy flow; adjusted watch paths to monitor deployment scripts and sample manifests.
  * Updated UI image build to load instrumented images without mutating manifests and altered E2E setup to scale/pause the operator and patch the UI proxy to use the local instrumented image.

* **Bug Fixes / Behavior**
  * Removed an extra post-login “Grant Access” interaction to improve E2E reliability.
  * Set local test namespace default to "default-tenant" for consistent local runs.

### DIFF
--- a/.github/actions/konflux-ui-deploy-test/action.yaml
+++ b/.github/actions/konflux-ui-deploy-test/action.yaml
@@ -16,14 +16,12 @@ runs:
         tool-cache: false
         docker-images: false
 
-    # konflux-ci is cloned to /home/runner/work/konflux-ui/konflux-ui
-    # as it have to be at home folder to use helm/kind-action@v1 action
     - name: Checkout Repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       with:
         repository: 'konflux-ci/konflux-ci'
         path: 'konflux-ci'
-        ref: 'legacy'
+        ref: 'main'
 
     - name: Disable AppArmor
       # works around a change in ubuntu 24.04 that restricts Linux namespace access
@@ -31,41 +29,36 @@ runs:
       run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
       shell: bash
 
-    - name: Create k8s Kind Cluster
-      uses: helm/kind-action@v1
+    - name: Install Kind
+      uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc # v1
       with:
-        config: ./konflux-ci/kind-config.yaml
-        cluster_name: konflux
+        install_only: true
 
     - name: Show version information
       run: |
         : Show version information
-        kubectl version
+        kubectl version --client
         kind version
       shell: bash
 
+    - name: Deploy Konflux
+      env:
+        GITHUB_APP_ID: ${{ env.GH_APP_ID }}
+        GITHUB_PRIVATE_KEY: ${{ env.GH_APP_PRIVATE_KEY }}
+        WEBHOOK_SECRET: ${{ env.GH_APP_WEBHOOK_SECRET }}
+        QUAY_TOKEN: ${{ env.QUAY_TOKEN_TEST }}
+        QUAY_ORGANIZATION: ${{ env.QUAY_TEST_ORG }}
+        SMEE_CHANNEL: ${{ env.SMEE_URL }}
+      run: |
+        : Deploy Konflux using operator-based deployment
+        cd konflux-ci
+        ./scripts/deploy-local.sh
+      shell: bash
+
     - name: List namespaces
       run: |
         : List namespaces
         kubectl get namespace
-      shell: bash
-
-    - name: Deploying Dependencies
-      run: |
-        : Deploying Dependencies
-        ./konflux-ci/deploy-deps.sh
-      shell: bash
-
-    - name: List namespaces
-      run: |
-        : List namespaces
-        kubectl get namespace
-      shell: bash
-
-    - name: Wait for the dependencies to be ready
-      run: |
-        : Wait for the dependencies to be ready
-        ./konflux-ci/wait-for-all.sh
       shell: bash
 
     - name: Build konflux-ui image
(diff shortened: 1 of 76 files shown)


.github/actions/konflux-ui-deploy-test/action.yaml
```diff
@@ -16,44 +16,43 @@ runs:
         tool-cache: false
         docker-images: false
 
-    # konflux-ci is cloned to /home/runner/work/konflux-ui/konflux-ui
-    # as it have to be at home folder to use helm/kind-action@v1 action
     - name: Checkout Repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```
Katka92: Is there a reason for a specific sha instead of a tag?

marcin-michal: With supply chain attacks being more and more common, I thought it could be more secure to pin to a specific commit instead of using a mutable tag.

marcin-michal:
> ## Review: #829
> 
> **Head SHA:** [3c53c61](https://github.com/konflux-ci/konflux-ui/commit/3c53c615d93fff6cde4a978575278efb38f90f86) **Timestamp:** 2026-05-04T00:00:00Z **Outcome:** comment-only
> ### Summary
> 
> This PR cleanly migrates from the legacy multi-step Konflux CI deployment to the new operator-based `deploy-local.sh` approach. The action consolidates five separate deployment steps into one script invocation and injects the custom UI image via a strategic-merge kubectl patch instead of pre-build kustomize manipulation. E2E test changes (namespace default, login flow simplification) are consistent with the operator deployment's conventions. No critical or high issues found; one medium finding regarding a hardcoded Kind cluster name warrants verification.
> ### Findings
> #### Medium
> 
>     * **[Correctness]** `pr_check.sh:35` — `kind load image-archive konflux-ui.tar -n konflux` hardcodes the cluster name `konflux`. The old action explicitly created the cluster with `cluster_name: konflux`, but the new flow delegates cluster creation to `deploy-local.sh`. If that script uses a different cluster name, image loading will fail and E2E tests will run against the default (non-instrumented) UI image.
>       Remediation: Verify that `deploy-local.sh` creates the cluster with name `konflux`, or parameterize the cluster name in `pr_check.sh` (e.g., read from an env var or the deploy script output).
> 
> 
> #### Low
> 
>     * **[Correctness]** `.github/actions/konflux-ui-deploy-test/action.yaml` — The `kubectl wait --for=delete pod -l control-plane=controller-manager -n konflux-operator --timeout=2m` step will succeed immediately if no matching pods exist (e.g., operator never started). This is likely fine for the expected flow, but if the operator scale-down did not take effect, the subsequent patch could race with the operator reconciling the proxy deployment back to its original state. Consider adding a brief verification that the operator pod is actually gone before patching.
> 
> 
> #### Info
> 
>     * **[Style]** `.github/workflows/periodic-watch-changes.yaml:22` — `actions/checkout@v4` remains unpinned to a SHA, while the deploy action now pins both `actions/checkout` and `helm/kind-action` to commit SHAs. This is a pre-existing inconsistency, not introduced by this PR, but worth addressing for supply-chain consistency.
> 
> 
> ### Footer
> 
> Outcome: comment-only This review applies to SHA `3c53c615d93fff6cde4a978575278efb38f90f86`. Any push to the PR head clears this review and requires a new evaluation.

- validated that the cluster name is `konflux`

sahil143 (commented):
Do we need this change since we are moving our jobs to konflux? Is it worth having it?

marcin-michal:
> Do we need this change since we are moving our jobs to konflux? Is it worth having it?

In the mean time, I think it is. I talked with Abhi and from what I saw, the way konflux-ci is deployed using the old scripts, and not using the new operator approach. So it is also possible that this can be reused in the future.

StanislavJochman (dismissed):
LGTM

rrosatti (dismissed):
LGTM

e2e-tests/tests/basic-happy-path.spec.ts
```diff
@@ -172,7 +182,7 @@ describe('Basic Happy Path', () => {
 
     it('Verify Enterprise contract Test pipeline run Details', () => {
       UIhelper.clickRowCellInTable('Pipeline run List', 'Test', `${applicationName}-`);
-      DetailsTab.waitForPLRAndDownloadAllLogs(false);
+      DetailsTab.waitForPLRAndDownloadAllLogs(false, '(Succeeded|Failed)');
```
Katka92: I believe we want to check that the pipeline succeeded. If the pipeline Failed, we want to fail the test too.

marcin-michal: is it required for the EC tests to pass for this test? I switched it to allow failed as well, to make the e2e tests more resilient, as I believe the ECs shouldn't affect the rest of the checks

rrosatti: @marcin-michal did you notice any issues with the EC check in the e2e tests from this branch?

marcin-michal: @rrosatti  yes, afaik in the legacy deployment some the EC checks were skipped by default, which is not the case in the new deployment [[1]](https://github.com/konflux-ci/konflux-ci/blob/main/scripts/deploy-local.sh#L157). so the EC pipeline would always fail, as the local deployment couldn't pass all the security checks

rrosatti: maybe then we can allow checking for `'(Succeeded|Failed)'` only when e2e tests are running within local deployment? 🤔  cc: @Katka92

Katka92: Hm, ok, let's keep the check also for Failed but make it optional just for the local deployment and let's check just for Succeeded if the backend is running on stage.

Katka92 (dismissed):

rrosatti (dismissed):
LGTM. nice job! :)

rrosatti (dismissed):
LGTM

rrosatti (dismissed):

e2e-tests/cypress.config.ts
```diff
@@ -119,7 +119,7 @@ export default defineConfig({
         LOCAL_CLUSTER: false,
         LOGIN_PROVIDER: '',
         PERIODIC_RUN_STAGE: false,
-        PIPELINE: 'docker-build-oci-ta',
+        PIPELINE: 'docker-build-oci-ta-min',
```
Katka92, line 122: This will change the test flow from the default pipeline to the minimal pipeline. As E2E tests are supposed to test the "default" golden path scenario, I want to confirm - do we want this change?

rrosatti: hmm, that's a good catch 🤔 @marcin-michal do you remember if this change was strictly necessary for the e2e tests to pass? or maybe it's a leftover change?

marcin-michal: @Katka92 @rrosatti  I tried switching to `docker-build-oci-ta-min` as this pipeline should be faster and lighter, since it has less tasks to run (I was hoping this would help with the e2e flakiness). I don't think it should really changes the test behavior or coverage from the UI side, just less tasks would be rendered in the task list or the graph. Let me know if that's fine, or I should change it back to the non-min version.

Katka92: I feel like we're changing the focus from the product "happy path" to just "ui" happy path. And probably it's not a bad approach given the other teams have a higher focus on quality too, so we should experience fewer non-UI-related issues. Let's keep the min pipeline and monitor the stability of the tests.

Katka92 (approved):

rrosatti (approved):

PR #838, opened 2026-04-27
refactor(KFLUXUI-1216): add page reload for e2e pipeline run status check

rrosatti:
## Fixes
Fixes: https://redhat.atlassian.net/browse/KFLUXUI-1216

## Description
Refactors the `waitUntilStatusIsNotRunning` method in `PipelinerunsTabPage` to handle websocket connection loss during long pipeline runs on the local Konflux backend. Instead of a single 50-minute wait that can stall when the websocket drops, the method now waits up to 30 minutes with periodic checks, then reloads the page every ~3 minutes (up to 7 retries) to re-establish the connection and refresh the status.

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes
- [x] Other (please describe): e2e test improvement

## How to test or reproduce?
- Run the e2e tests that use `DetailsTab.waitUntilStatusIsNotRunning()` against a local Konflux backend
- Verify that when a pipeline run takes longer than 30 minutes, the page reloads and the status is re-checked
- Verify that if the pipeline completes within 30 minutes, no reload occurs

## Browser conformance:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test reliability for monitoring pipeline run status with configurable timeout, polling interval, and retry behavior.
  * Tests now poll status text, auto-refresh, and retry when a run remains in progress to reduce flakiness and catch stalled runs.
  * Tests explicitly fail after maximum retries to surface persistent issues promptly.

### DIFF
--- a/e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
+++ b/e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
@@ -105,12 +105,42 @@ export class DetailsTab {
     UIhelper.clickTab('Details');
   }
 
-  // Setting timeout to 50 minutes (Builds sometimes taking slightly longer than 40 minutes).
-  static waitUntilStatusIsNotRunning() {
-    cy.get(pipelinerunsTabPO.statusPO).scrollIntoView();
-    cy.get(pipelinerunsTabPO.statusPO, { timeout: 3000000 })
-      .should('not.have.text', 'Pending')
-      .and('not.have.text', 'Running');
+  // wait for status to be not running, if it is still running after retries, throw error to fail the test
+  static waitUntilStatusIsNotRunning(
+    timeoutDuration: number = 1800000, // 30 minutes
+    interval: number = 60000, // 60 seconds
+    retries: number = 7, // number of retries before failing the test
+  ) {
+    if (retries <= 0) {
+      assert.fail('Pipelinerun is still running after maximum retries');
+    }
+
+    const maxAttempts = Math.max(1, Math.floor(timeoutDuration / interval));
+
+    const checkStatus = (attempt: number) => {
+      cy.get(pipelinerunsTabPO.statusPO)
+        .scrollIntoView()
+        .invoke('text')
+        .then((statusText) => {
+          if (!statusText.includes('Pending') && !statusText.includes('Running')) {
+            expect(statusText).to.not.include('Running');
+            return;
+          }
+          if (attempt >= maxAttempts) {
+            cy.log(
+              `Status still running. Reloading page (${retries - 1} ${retries - 1 === 1 ? 'retry' : 'retries'} remaining).`,
+            );
+            cy.reload();
+            // 180000 ms = 3 minutes
+            this.waitUntilStatusIsNotRunning(180000, interval, retries - 1);
+            return;
+          }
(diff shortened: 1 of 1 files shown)


e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
```diff
@@ -105,12 +105,42 @@ export class DetailsTab {
...
+
+    const checkStatus = (attempt: number) => {
+      cy.get(pipelinerunsTabPO.statusPO)
+        .scrollIntoView()
+        .invoke('text')
+        .then((statusText) => {
+          if (!statusText.includes('Pending') && !statusText.includes('Running')) {
+            expect(statusText).to.not.include('Running');
+            return;
+          }
```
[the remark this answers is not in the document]
Katka92: I agree with coderabbit there - this expect is redundant.

e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
```diff
@@ -105,12 +105,42 @@ export class DetailsTab {
...
+          if (attempt >= maxAttempts) {
+            cy.log(
+              `Status still running. Reloading page (${retries - 1} ${retries - 1 === 1 ? 'retry' : 'retries'} remaining).`,
```
Katka92: At this point, we're not sure what status actually is - consider printing the actual status (may be running or pending) or rephrasing the log.

e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
```diff
@@ -105,12 +105,42 @@ export class DetailsTab {
...
+            );
+            cy.reload();
+            // 180000 ms = 3 minutes
+            this.waitUntilStatusIsNotRunning(180000, interval, retries - 1);
+            return;
+          }
+          cy.wait(interval);
```
Katka92, line 141: I don't think we need to check this interval because Cypress automatically watch the page and checks for the element. I would maybe go with the previous code (with a timeout of 30 mins), but then catch the error and do this reload logic instead of failing the test. WDYT?

Katka92 (changes requested):

e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
rrosatti: As I pointed via Slack, I was not able to follow this approach. Unless I misunderstood something :)

Katka92: Yeah, you're right, we can't catch the error easily there. I'm sorry for the confusion. I'm not a big fan of recursions as they seem harder to understand, but let's keep it as it is. Would you mind adding a short comment explaining the logic in this method and the reasoning for the recursion? Thank you

rrosatti: done! I added a comment on top of the function with a better explanation :)

JoaoPedroPP (approved):
LGTM

Tested locally E2E worked.

Katka92 (approved):

PR #852, opened 2026-05-03
feat(KFLUXSE-166): add Conforma Policy Status Overview page (PoC)

Omeramsc:
## Fixes 
Fixes: https://issues.redhat.com/browse/KFLUXSE-166

## Description
Implements an end-to-end proof-of-concept for the workspace-level Policy page that aggregates Conforma/EC results across all components. Gated behind the `conforma-policy` feature flag (wip, no guard).

New files:
- src/components/Policy/PolicyPage.tsx — page shell with summary header
- src/components/Policy/PolicyConformaTable.tsx — flat results table
- src/components/Policy/usePolicyConformaResults.ts — data aggregation hook
- src/routes/page-routes/policy.tsx — lazy route definition

Modified:
- Feature flag registered, route path added, sidebar nav item added.

* **New Features**
  * Added Conforma Results tab to application details, displaying security scan results with search, status filtering, and rule/component grouping options.
  * Summary metrics show violation, warning, and success counts per component.
  * Expandable result groups reveal detailed information including descriptions, images, and remediation guidance.

## Type of change
- [x] Feature

Screenshot 2026-06-04 at 1 47 01
Screenshot 2026-06-04 at 1 46 00
Screenshot 2026-06-04 at 1 46 12
Screenshot 2026-06-04 at 1 46 21
Screenshot 2026-06-04 at 1 46 32
Screenshot 2026-06-04 at 1 46 41
Screenshot 2026-06-04 at 1 46 52

### Unit tests
```bash
yarn test --testPathPattern='Conforma|Issues'
```

### Manual testing — mock data (no cluster needed)

1. Start the dev server: yarn start
2. Navigate to any application's detail page and append ?mock=conforma to the URL
- e.g. https://localhost:8080/ns/<workspace>/applications/<app>/conforma-results?mock=conforma
3. Verify the Conforma Results tab renders:
- Summary bar with tooltips (hover each metric)
- Grouped expandable table (toggle "Group by: Rule" / "Group by: Component")
- Search, status filter, expand all / collapse all

### Manual testing — real cluster
1. Enable the conforma-policy feature flag in the Experimental Features panel
2. Open an application that has pipeline runs with verify-enterprise-contract or conforma-policy tasks
3. Click the Conforma Results tab and verify results load
4. On the Issues Dashboard, verify the violations card aggregates across all applications in the workspace

What to verify:
 [ ] Conforma Results tab visible on Application Details (even when empty)
 [ ] Summary bar shows correct counts with vertical dividers and hover tooltips
 [ ] Count badges use PatternFly Label outline style (pill shape with colored border)
 [ ] Group by rule / component toggles table structure correctly
 [ ] Search and status filters work
 [ ] Expand all / Collapse all buttons work
 [ ] Empty state message when no results
 [ ] Loading spinner while fetching
 [ ] Error state when fetch fails
 [ ] Issues Dashboard shows "Conforma Policy Violations" card (behind feature flag)
 [ ] "View policy" links navigate to the correct application's Conforma Results tab

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## New Features
* Added "Conforma Results" tab to application details, displaying security compliance results with violations, warnings, and successes
* Filter results by name search and status
* Group results by rule or component with expandable rows for detailed information
* Summary metrics display total violations, warnings, and successful checks across all components


src/routes/index.tsx
```diff
@@ -15,6 +15,7 @@ import integrationTestRoutes from './page-routes/integration-test';
 import issuesRoutes from './page-routes/issues';
 import workspaceRoutes from './page-routes/namespace';
 import pipelineRoutes from './page-routes/pipeline';
+import policyRoutes from './page-routes/policy';
```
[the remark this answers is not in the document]
Omeramsc: changing only the policy import to an alias while every sibling uses relative paths would be inconsistent

src/components/Policy/PolicyConformaTable.tsx
JoaoPedroPP, line 1: I think this should follow the same pattern we currently have for other table components like `CommitsListViewV2.tsx`, `CommitsListRow.tsx` and `CommitsListHeader.tsx`. These components make use of `Table` component and shared component `TableData` component that abstract how to build tables and tabla data.

JoaoPedroPP (changes requested):
This is a nice update but I think we should follow the current pattern we currently have. Also please take a look at bot comments and add unit tests to assure maximum coverage of components and hit the minimum coverage required.

src/components/Policy/PolicyConformaTable.tsx
StanislavJochman: +1

src/components/ApplicationDetails/ApplicationDetails.tsx
```diff
@@ -193,6 +196,16 @@ export const ApplicationDetails: React.FC<React.PropsWithChildren> = () => {
...
+          ...(isConformaPolicyEnabled
+            ? [
+                {
+                  key: 'conforma-results',
+                  label: 'Conforma Results',
+                  isFilled: true,
+                  featureFlag: ['conforma-policy'] as FlagKey[],
+                },
+              ]
+            : []),
```
JoaoPedroPP, line 215: I don't think this aligned to the rest of options. Why is in this case an array while for other items is an object?

Image

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const [searchText, setSearchText] = React.useState('');
+  const [groupBy, setGroupBy] = React.useState<GroupByMode>('rule');
+  const [statusFilters, setStatusFilters] = React.useState<string[]>([]);
+  const [expandedGroups, setExpandedGroups] = React.useState<Set<string>>(new Set());
+
+  const filteredResults = React.useMemo(
+    () => filterResults(allResults, searchText, statusFilters),
+    [allResults, searchText, statusFilters],
+  );
```
JoaoPedroPP, line 68: We have [guidance](https://github.com/Omeramsc/konflux-ui/blob/feat/conforma-policy-poc/docs/generic-filter-system.md#before-current-pipelinerunlistview-implementation) for how filters should work. Currently is required to use `createFilterObj()` function.

src/components/Conforma/ConformaResultsTab/ConformaSummaryBar.tsx
JoaoPedroPP, line 1: I am not sure this is working. I have six component which only 3 are fully onboarded but when I group by components I can only see 2 components.

Image

JoaoPedroPP (changes requested):
Overall the new dashboard is working with some details not working as expected, which I left some comments below.

There is a problem in this PR that I see a lot of new components that create tables that are not following the current design pattern. When creating a new table we ask to follow the current design pattern implemented in components like PipelinesRun or Commits so we have common way to create tables and filters. We have this [guidance](https://github.com/Omeramsc/konflux-ui/blob/feat/conforma-policy-poc/docs/guidelines/table-component.md) explaining how to implement table components. Have you try to use it? Is there any feature that you required that is not covered by the component?

I notice in this new tab the design is slightly different from other table components we currently have (I think similar table would be the security tab in detailed pipelinerun, see image below)  so I need to ask, is this design approved? Konflux UI designer is pankaj.

Image

Omeramsc:
> Overall the new dashboard is working with some details not working as expected, which I left some comments below.
> 
> There is a problem in this PR that I see a lot of new components that create tables that are not following the current design pattern. When creating a new table we ask to follow the current design pattern implemented in components like PipelinesRun or Commits so we have common way to create tables and filters. We have this [guidance](https://github.com/Omeramsc/konflux-ui/blob/feat/conforma-policy-poc/docs/guidelines/table-component.md) explaining how to implement table components. Have you try to use it? Is there any feature that you required that is not covered by the component?
> 
> I notice in this new tab the design is slightly different from other table components we currently have (I think similar table would be the security tab in detailed pipelinerun, see image below) so I need to ask, is this design approved? Konflux UI designer is pankaj.
> 
> Image

Thanks for the review, I'll look into the specific guidance against what was created. 
The design was made by pankaj and is detailed here:  https://redhat.atlassian.net/browse/DTUX-2774

src/components/Conforma/ConformaResultsTab/ConformaSummaryBar.tsx
Omeramsc: does the third component have pipelineruns? only components with at least one Conforma result row (violation, warning, or success) appear; components with no fetched Conforma data are counted in the summary but omitted from the grouped table

Omeramsc:
> Overall the new dashboard is working with some details not working as expected, which I left some comments below.
> 
> There is a problem in this PR that I see a lot of new components that create tables that are not following the current design pattern. When creating a new table we ask to follow the current design pattern implemented in components like PipelinesRun or Commits so we have common way to create tables and filters. We have this [guidance](https://github.com/Omeramsc/konflux-ui/blob/feat/conforma-policy-poc/docs/guidelines/table-component.md) explaining how to implement table components. Have you try to use it? Is there any feature that you required that is not covered by the component?
> 
> I notice in this new tab the design is slightly different from other table components we currently have (I think similar table would be the security tab in detailed pipelinerun, see image below) so I need to ask, is this design approved? Konflux UI designer is pankaj.
> 
> Image

agreed on aligning with the established table infrastructure. I just refactored it:

- Added ConformaResultsListHeader.ts using createTableHeaders
- Added ConformaResultsListRow.tsx using the Row fragment + TableData pattern
- ConformaGroupedTable now uses those shared pieces for column config and cell rendering
- One intentional deviation: we kept the raw PatternFly expandable Table rather than ~/shared Table because this view needs external Expand all / Collapse all control. The shared Table's expand prop manages expansion state internally, which conflicted with the toolbar buttons. The header/row/cell layer now matches ConformaTable / CommitsListViewV2; only the expandable shell differs.

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
Omeramsc: Fixed, and now following the same pattern as SecurityConformaTab:

src/components/ApplicationDetails/ApplicationDetails.tsx
Omeramsc: I removed the prop rather than normalizing it, since conditional rendering already handles visibility.

src/components/Conforma/ConformaResultsTab/__tests__/ConformaResultsToolbar.spec.tsx
```diff
@@ -0,0 +1,128 @@
+import { fireEvent, screen } from '@testing-library/react';
+import { FilterContextProvider } from '~/components/Filter/generic/FilterContext';
+import { CONFORMA_RESULT_STATUS } from '~/types/conforma';
+import { routerRenderer } from '~/unit-test-utils/mock-react-router';
+import type { GroupByMode } from '../conforma-grouping-utils';
+import { ConformaResultsToolbar } from '../ConformaResultsToolbar';
+import type { ConformaResultRow } from '../useApplicationConformaResults';
```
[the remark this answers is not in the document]
Omeramsc: The project convention in docs/best-practices.md forbids deep cross-directory relative paths (e.g. ../../../types), not intra-module ../ references from spec files into their parent directory. 
All 13 spec files in this module use the same ../ pattern,  as do spec files throughout the codebase. 
Changing only these three imports without a codebase-wide migration would create an inconsistency rather than resolve one.

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
```diff
@@ -0,0 +1,114 @@
...
+  SelectOption,
+} from '@patternfly/react-core';
+import { FilterContext } from '~/components/Filter/generic/FilterContext';
+import { MultiSelect } from '~/components/Filter/generic/MultiSelect';
+import { BaseTextFilterToolbar } from '~/components/Filter/toolbars/BaseTextFIlterToolbar';
+import { createFilterObj } from '~/components/Filter/utils/filter-utils';
+import { useDeepCompareMemoize } from '~/shared';
+import { CONFORMA_RESULT_STATUS } from '~/types/conforma';
+import type { GroupByMode } from './conforma-grouping-utils';
+import type { ConformaResultRow } from './useApplicationConformaResults';
```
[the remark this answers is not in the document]
Omeramsc: same.

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
```diff
@@ -0,0 +1,114 @@
...
+  onGroupByChange,
+  onExpandAll,
+  onCollapseAll,
+}) => {
+  const { filters: unparsedFilters, setFilters, onClearFilters } = React.useContext(FilterContext);
+  const filters = useDeepCompareMemoize({
+    name: unparsedFilters.name ? (unparsedFilters.name as string) : '',
+    status: unparsedFilters.status ? (unparsedFilters.status as string[]) : [],
+  });
+  const { name: nameFilter, status: statusFilter } = filters;
```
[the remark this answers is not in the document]
Omeramsc: same reasoning as above.

src/components/Conforma/ConformaResultsTab/ConformaResultsListRow.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import { TableData } from '~/shared';
+import type { GroupedConformaRow } from './conforma-grouping-utils';
+import { ConformaCountBadge } from './ConformaCountBadge';
+import { ConformaGroupedTableColumnClasses } from './ConformaResultsListHeader';
```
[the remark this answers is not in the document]
Omeramsc: same

src/components/Conforma/ConformaResultsTab/__tests__/ConformaSummaryBar.spec.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  it('renders dividers between items', () => {
+    const { container } = routerRenderer(<ConformaSummaryBar {...defaultProps} />);
+    const dividers = container.querySelectorAll('[class*="divider"]');
+    expect(dividers.length).toBe(4);
+  });
+
+  it('has the correct data-test attribute', () => {
+    routerRenderer(<ConformaSummaryBar {...defaultProps} />);
+    expect(screen.getByTestId('conforma-summary-bar')).toBeInTheDocument();
```
JoaoPedroPP: Instead of `getByTestId` can we have visual test case like `getByText` or something else?

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
JoaoPedroPP, line 1: I have the impression this file condense a lot of information in one single place. There are hooks, types and queries here that I think could be moved to specific directories like hooks and types. @sahil143 What do you think?

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -0,0 +1,552 @@
...
+type QualifyingPipeline = {
+  componentName: string;
+  pipelineRun: PipelineRunKind;
+  securityTaskName: typeof EC_TASK | typeof CONFORMA_TASK;
+};
+
+type ConformaByComponentEntry =
+  | { state: 'ok'; components: ComponentConformaResult[]; pipelineRunName: string }
+  | { state: 'error'; err: unknown; pipelineRunName?: string }
+  | { state: 'skipped' };
```
JoaoPedroPP: I think this should be moved to `/types/` directory

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -0,0 +1,552 @@
...
+
+  return rows;
+};
+
+export const useApplicationConformaResults = (
+  applicationName: string,
+): ApplicationConformaResults => {
+  const { search } = useLocation();
+  const useMock =
+    process.env.NODE_ENV === 'development' && new URLSearchParams(search).has('mock', 'conforma');
```
JoaoPedroPP: Should this be here?

src/components/Conforma/ConformaResultsTab/ConformaSummaryBar.tsx
JoaoPedroPP: Good point, I think it failed in Conforma pipelinerun

e2e-tests/tests/basic-happy-path.spec.ts
```diff
@@ -178,13 +178,7 @@ describe('Basic Happy Path', () => {
...
     it('Verify Enterprise contract Test pipeline run Details', () => {
       UIhelper.clickRowCellInTable('Pipeline run List', 'Test', `${applicationName}-`);
-      // We encountered problems with EC checks on a local deployment,
-      // so we only check for Succeeded status on the stage job
-      if (Cypress.env('PERIODIC_RUN_STAGE') === 'true') {
-        DetailsTab.waitForPLRAndDownloadAllLogs(false, 'Succeeded');
-      } else {
-        DetailsTab.waitForPLRAndDownloadAllLogs(false, '(Succeeded|Failed)');
-      }
+      DetailsTab.waitForPLRAndDownloadAllLogs(false);
```
JoaoPedroPP: @Katka92 Is this change ok?

e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
```diff
@@ -205,10 +205,10 @@ export class DetailsTab {
       });
   }
 
-  static waitForPLRAndDownloadAllLogs(allTaskLogs = true, expectedStatus = 'Succeeded') {
```
JoaoPedroPP, line 208: Why remove this?

JoaoPedroPP (commented):
Overall it is working as expected. Left some comments about small things. Nice job

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
Omeramsc: On a second look, it is. There are some parts of it that can def be separated. i'll take the 3 exported types, as well as listSecurityTaskRuns, resolveConformaResultFromTaskRun, fetchConformaForPipeline, to a separate file. 
If further separation is needed, lmk

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
Omeramsc: IMO those should live in conforma.ts, as they are conforma specifics and not They are not hook-specific internals . moved `ConformaResultRow`, `ComponentConformaStatus`, and `ApplicationConformaResults` to `src/types/conforma.ts` and re-import them.

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
Omeramsc: Right, extracting it to conforma-fetchers.ts

src/components/Conforma/ConformaResultsTab/__tests__/ConformaSummaryBar.spec.tsx
Omeramsc: Correct, it doesn't really cover much. will change

e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
Omeramsc: Since the UI does not self-refresh, there is flackiness this is trying to solve. the solution may be a bit sub-optimal, I've refined it a bit. WDYT?

src/routes/page-routes/application.tsx
```diff
@@ -3,6 +3,7 @@ import { ActivityTab } from '~/components/Activity';
 import { ApplicationDetails, ApplicationOverviewTab } from '~/components/ApplicationDetails';
 import { applicationPageLoader, ApplicationListView } from '~/components/Applications';
 import { ComponentListTab, componentsTabLoader } from '~/components/Components/ComponentsListView';
+import { ConformaResultsTab } from '~/components/Conforma/ConformaResultsTab/ConformaResultsTab';
 import { FilterContextProvider } from '~/components/Filter/generic/FilterContext';
```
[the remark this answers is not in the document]
Omeramsc: nit, out of current scope, will not fix.

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  const handleCollapseAll = React.useCallback(() => {
+    setExpandedGroups(new Set());
+  }, []);
+
+  if (error) {
+    return (
+      <PageSection>
+        <Text>Unable to load Conforma results.</Text>
+      </PageSection>
```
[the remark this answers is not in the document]
Omeramsc: nit, out of current scope, will not fix.

e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
rrosatti, line 108: these e2e changes shouldn't be here, you need to rebase your branch with main :)

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -0,0 +1,411 @@
...
+  }, [useMock]);
+
+  const [conformaByComponent, setConformaByComponent] = React.useState<
+    Record<string, ConformaByComponentEntry | undefined>
+  >({});
+  const [conformaBatchFinished, setConformaBatchFinished] = React.useState(false);
+
+  const readyToFetch = Boolean(namespace?.length && pipelinesLoaded && componentsLoaded);
+
+  React.useEffect(() => {
```
rrosatti: thinking out loud: maybe we could explore using TanStack useQueries hook here? https://tanstack.com/query/v4/docs/framework/react/reference/useQueries. It might simplify some things. What do you guys think? 🤔

src/components/Conforma/ConformaResultsTab/ConformaGroupedTable.tsx
```diff
@@ -0,0 +1,125 @@
...
+              </div>
+            )}
+          </Td>
+        </Tr>
+      ))}
+    </Tbody>
+  </Table>
+);
+
+export const ConformaGroupedTable: React.FC<ConformaGroupedTableProps> = ({
```
rrosatti, line 66: yesterday Sahil shared an amazing TableV2 component in PR #997. I'm wondering if it could be used here instead of creating a new/custom Table component 

(apologies if I'm repeating some discussion that might already have happened before about it - if so, just ignore me :D)

cc: @sahil143

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const handleExpandAll = React.useCallback(() => {
+    setExpandedGroups(new Set(groups.map((g) => g.groupKey)));
+  }, [groups]);
+
+  const handleCollapseAll = React.useCallback(() => {
+    setExpandedGroups(new Set());
+  }, []);
+
+  if (error) {
```
rrosatti: nit: can we maybe use the `getErrorState(..)` utility here? We're using in quite a few places already, so it should be easy to find examples :)

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -0,0 +1,411 @@
...
+    let cancelled = false;
+    void import('./__data__/mockConformaResults')
+      .then(({ generateMockResults }) => {
+        if (!cancelled) setMockResults(generateMockResults());
+      })
+      .catch((err) => {
+        logger.warn('Failed to load mock Conforma data', { error: err });
+        if (!cancelled) {
+          setMockResults({
+            componentStatuses: [],
```
rrosatti: it seems this "default empty object" is repeated 3 times in the file, we could extract it to a constant:

```tsx
  const EMPTY_RESULTS: ApplicationConformaResults = {
    componentStatuses: [],
    allResults: [],
    totalComponents: 0,
    totalFailed: 0,
    totalViolations: 0,
    totalWarnings: 0,
    totalSuccesses: 0,
    loaded: false,
    error: undefined,
  };
``` 

and use it where needed :)

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -0,0 +1,411 @@
...
+  );
+
+  const batchError = React.useMemo(() => {
+    const entries = Object.values(conformaByComponent);
+    const errorEntries = entries.filter((v) => v?.state === 'error');
+    const allFailed = entries.length > 0 && entries.every((v) => v?.state === 'error');
+    return allFailed && errorEntries[0]?.state === 'error' ? errorEntries[0].err : undefined;
+  }, [conformaByComponent]);
+
+  React.useEffect(() => {
```
rrosatti: hmm, do we really need this extra `useEffect` just to log warn the errors? 🤔 maybe we can call `logger.warn()` inside the `runFetch` in the catch logic (e.g. line 268)

Omeramsc: Added logger.warn('Conforma fetch failed for component', { component, pipelineRun, error }) directly in the catch of runFetch, abd Removed the useEffect that scanned conformaByComponent for errors and logged them

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
Omeramsc: Extracted a single EMPTY_RESULTS: ApplicationConformaResults and reused it in three places

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
Omeramsc: done

e2e-tests/support/pages/tabs/PipelinerunsTabPage.ts
Omeramsc: correct, rebased

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
Omeramsc: I think we may follow-up with it later if needed. currently useEffect and Promise.all are working and tested, refactoring our of scope will delay this longer.

src/components/Conforma/ConformaResultsTab/ConformaGroupedTable.tsx
Omeramsc: Since this is still a draft, we cannot depend on it without blocking ourselvs. we are aligned with the guidelines, I'll suggest we'll keep it as is and consider a refactor later.

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
```diff
@@ -0,0 +1,127 @@
...
+import { TektonResourceLabel } from '~/types/coreTekton';
+import { getPipelineRunFromTaskRunOwnerRef } from '~/utils/common-utils';
+import { getTaskRunLog } from '~/utils/tekton-results';
+
+async function listSecurityTaskRuns(
+  ns: string,
+  pipelineRunName: string,
+  pipelineTaskName: string,
+): Promise<TaskRunKind[]> {
+  return K8sListResourceItems<TaskRunKind>({
```
sahil143: This should use `K8sQueryListResourceItems` to fetch from archive also

sahil143 (commented):
@Omeramsc I am not seeing any data on the Conforma page.
Image

It's because you are not fetching taskruns from archive but only from cluster!

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -0,0 +1,337 @@
...
+      if (!components) {
+        return {
+          componentName: name,
+          status: 'unknown' as const,
+          violationCount: 0,
+          warningCount: 0,
+          successCount: 0,
+          pipelineRunName,
+        };
+      }
```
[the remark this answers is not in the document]
Omeramsc: Added `isKubearchiveLogsEnabled` to the queryKey so React Query invalidates the cache when the flag changes

src/components/Conforma/ConformaResultsTab/conforma-grouping-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+  const lowerSearch = searchText.toLowerCase();
+
+  return results.filter((row) => {
+    if (
+      lowerSearch &&
+      !row.title?.toLowerCase().includes(lowerSearch) &&
+      !row.component?.toLowerCase().includes(lowerSearch)
+    ) {
+      return false;
+    }
```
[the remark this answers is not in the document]
Omeramsc: Replaced the manual matching with textMatch, removing the pre-lowercased variable and simplifying the function

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  if (podLogOpts) {
+    try {
+      return await commonFetchJSON<ConformaResult>(
+        getK8sResourceURL(PodModel, undefined, podLogOpts),
+      );
+    } catch (err) {
+      const code =
+        typeof err === 'object' && err !== null && 'code' in err && typeof err.code === 'number'
+          ? err.code
```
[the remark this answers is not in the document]
Omeramsc: Added a logger.debug call for the 404-without-kubearchive case so the fallback path is observable in debug logs.

src/components/Conforma/ConformaResultsTab/conforma-grouping-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+  if (allComponentNames) {
+    for (const name of allComponentNames) {
+      if (name) map.set(name, []);
+    }
+  }
+
+  for (const row of results) {
+    const key = row.component || 'Unknown component';
+    const existing = map.get(key);
+    if (existing) {
```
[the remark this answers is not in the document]
Omeramsc: rules returned by Conforma always have titles.

src/components/ApplicationDetails/ApplicationDetails.tsx
```diff
@@ -193,6 +195,15 @@ export const ApplicationDetails: React.FC<React.PropsWithChildren> = () => {
             label: 'Releases',
             isFilled: true,
           },
+          ...(isConformaPolicyEnabled
```
rrosatti, line 207: I noticed that there's no feature flag being displayed on the tab 🤔 

Image

rrosatti: we might need to do the following changes in order for it to be displayed:

```diff
diff --git a/src/components/ApplicationDetails/ApplicationDetails.tsx b/src/components/ApplicationDetails/ApplicationDetails.tsx
index 0acb94d9..28ca2033 100644
--- a/src/components/ApplicationDetails/ApplicationDetails.tsx
+++ b/src/components/ApplicationDetails/ApplicationDetails.tsx
@@ -4,6 +4,7 @@ import { Bullseye, Spinner } from '@patternfly/react-core';
 import { useIsOnFeatureFlag } from '~/feature-flags/hooks';
 import { useNamespace } from '~/shared/providers/Namespace';
 import { getErrorState } from '~/shared/utils/error-utils';
+import { FlagKey } from '../../feature-flags/flags';
 import { useApplication } from '../../hooks/useApplications';
 import { ApplicationModel, ComponentModel, IntegrationTestScenarioModel } from '../../models';
 import {
@@ -201,6 +202,7 @@ export const ApplicationDetails: React.FC<React.PropsWithChildren> = () => {
                   key: 'conforma-results',
                   label: 'Conforma Results',
                   isFilled: true,
+                  featureFlag: ['conforma-policy'] as FlagKey[],
                 },
               ]
             : []),
diff --git a/src/components/DetailsPage/types.ts b/src/components/DetailsPage/types.ts
index 1bd83b41..235deb6f 100644
--- a/src/components/DetailsPage/types.ts
+++ b/src/components/DetailsPage/types.ts
@@ -1,4 +1,5 @@
 import { DropdownItemProps } from '@patternfly/react-core';
+import { FlagKey } from '../../feature-flags/flags';

 export type Action = {
   type?: string;
@@ -16,4 +17,5 @@ export type DetailsPageTabProps = {
   partial?: boolean;
   className?: string;
   isFilled?: boolean;
+  featureFlag?: FlagKey[];
 };
```

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
```diff
@@ -0,0 +1,86 @@
...
+        queryParams: { container: 'step-report-json', follow: 'true' },
+      }
+    : null;
+
+  if (podLogOpts) {
+    try {
+      return await commonFetchJSON<ConformaResult>(
+        getK8sResourceURL(PodModel, undefined, podLogOpts),
+      );
+    } catch (err) {
```
rrosatti: can we somehow re-use the `fetchResourceWithK8sAndKubeArchive` utility here? 🤔

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
```diff
@@ -0,0 +1,86 @@
...
+    try {
+      const logs = await getTaskRunLog(taskRunNs, taskRunUid, pipelineRunUid);
+      return extractConformaResultsFromTaskRunLogs(logs);
+    } catch (e) {
+      logger.warn('Conforma aggregate: tekton-results log fallback failed', { error: e });
+    }
+  }
+
+  return undefined;
+}
```
rrosatti, line 97: I see that this function has some similarity with the logic in the `src/components/Conforma/useConformaResult.tsx` hook. I'm wondering whether we could make some of this logic re-usable in both places? 🤔

src/components/Conforma/ConformaResultsTab/ConformaCountBadge.tsx
```diff
@@ -0,0 +1,36 @@
...
+  type: 'violations' | 'warnings' | 'successes';
+};
+
+const badgeConfig = {
+  violations: { icon: <ExclamationCircleIcon />, color: 'red' as const },
+  warnings: { icon: <ExclamationTriangleIcon />, color: 'gold' as const },
+  successes: { icon: <CheckCircleIcon />, color: 'green' as const },
+};
+
+export const ConformaCountBadge: React.FC<ConformaCountBadgeProps> = ({ count, type }) => {
```
rrosatti, line 11: can we try to re-use the `getRuleStatus` function here? (`src/components/Conforma/utils.tsx`). It has the logic that takes care of returning the right icon for each "type"

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
```diff
@@ -0,0 +1,113 @@
...
+        onSelect={(_, value) => {
+          onGroupByChange(value as GroupByMode);
+          setIsGroupByOpen(false);
+        }}
+        selected={groupBy}
+        isOpen={isGroupByOpen}
+        onOpenChange={setIsGroupByOpen}
+      >
+        <SelectList>
+          <SelectOption value="rule">Group by: Rule</SelectOption>
```
rrosatti: nit: I'd say we can omit the "Group by: ". We're already  showing it in the `MenuToggle` text :)

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -0,0 +1,337 @@
...
+
+  return rows;
+};
+
+export const useApplicationConformaResults = (
+  applicationName: string,
+): ApplicationConformaResults => {
+  const { search } = useLocation();
+  const useMock =
+    process.env.NODE_ENV === 'development' && new URLSearchParams(search).has('mock', 'conforma');
```
rrosatti: I believe we should not include this in the code. If it's necessary to mock data to test the feature, then I'd say the PR description could be a good place to point that out 🤔 what do you guys think?

rrosatti (commented):
thanks for the updates! I left some comments, I'd suggest to wait for Sahil's expertise input here as well, since he'll def catch some details that need to be fixed/updated :)

src/components/ApplicationDetails/ApplicationDetails.tsx
sahil143: It's alright if there is no flag on the tab. Just having having it on the tab content should be fine

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
Omeramsc: Agreed, I entered it by mistake. will fix.

src/components/Conforma/ConformaResultsTab/ConformaCountBadge.tsx
Omeramsc: ConformaCountBadge now also reads from RULE_STATUS_CONFIG instead of maintaining a duplication.

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
Omeramsc: remoed the Group by, now showing only "Rule" / "Component"

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
Omeramsc: The two useEffect blocks in `useConformaResult.tsx` were deleted and replaced by a single effect that calls `resolveConformaResultFromTaskRun()` from `conforma-fetchers.ts`. does that align with the reusability?

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
Omeramsc: same as above

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
```diff
@@ -0,0 +1,135 @@
...
+  if (taskRunUid && taskRunNs && pipelineRunUid) {
+    try {
+      const logs = await getTaskRunLog(taskRunNs, taskRunUid, pipelineRunUid);
+      return extractConformaResultsFromTaskRunLogs(logs);
+    } catch (e) {
+      logger.warn('Conforma aggregate: tekton-results log fallback failed', { error: e });
+    }
+  }
+
+  return undefined;
```
sahil143: **Bug:** `resolveConformaResultFromTaskRun` catches every error internally and returns `undefined`, so the `queryFn` never throws. React Query always sets `status: 'success'` and `error: null`.

This means the consumer code in `useApplicationConformaResults.ts` that checks `q.isError`:

```ts
const allQueriesFailed =
  logQueries.length > 0 && logQueries.every((q) => q.isError); // always false

const aggregateError =
  componentsError ?? taskRunsError ?? (allQueriesFailed ? queryErrors[0] : undefined);
```

...is dead code. When all fetches fail, users see an empty results view instead of an error state.

The test `'sets error when all component queries fail'` mocks `mockResolveConforma.mockRejectedValue(...)`, but the real function never rejects -- it catches everything. So the test passes against a scenario that can't happen in production.

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -0,0 +1,250 @@
...
+    });
+
+    const mergedConformaRows: ComponentConformaResult[] = [];
+    for (const components of conformaByComponent.values()) {
+      mergedConformaRows.push(...components);
+    }
+
+    const allResults: ConformaResultRow[] = mapConformaResultData(mergedConformaRows);
+
+    const totalViolations = allResults.filter(
```
sahil143: **Nitpick:** Three `.filter().length` passes over the same array:

```ts
const totalViolations = allResults.filter(r => r.status === ...).length;
const totalWarnings = allResults.filter(r => r.status === ...).length;
const totalSuccesses = allResults.filter(r => r.status === ...).length;
```

A single `.reduce()` would do this in one pass. Same pattern in `groupByRule`/`groupByComponent` (3 `.filter().length` per group).

src/components/Conforma/ConformaResultsTab/__tests__/ConformaSummaryBar.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+    expect(screen.getByText('Failed Components')).toBeInTheDocument();
+    expect(screen.getByText('Violations')).toBeInTheDocument();
+    expect(screen.getByText('Warnings')).toBeInTheDocument();
+    expect(screen.getByText('Successes')).toBeInTheDocument();
+  });
+
+  it('displays correct counts for each metric', () => {
+    const { container } = routerRenderer(<ConformaSummaryBar {...defaultProps} />);
+
+    const strongElements = container.querySelectorAll('strong');
```
sahil143: **Nitpick:** `container.querySelectorAll('strong')` and `container.querySelectorAll('[class*="divider"]')` query by DOM structure. Project guidelines prefer semantic queries (`getByRole`, `getByText`, etc.).

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
Omeramsc: This is important, thanks for pointing it out. I've fixed  it so fetch failures throw after all fallbacks are exhausted, and make sure the test follow the production path.
in addition, I want to Distinguish "no data" from "fetch failed" in the UI. Needsto think about the UX on how to display it (badge? tooltip?) but either way, out of scope of this change

src/components/Conforma/ConformaResultsTab/__tests__/ConformaSummaryBar.spec.tsx
Omeramsc: fixed

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
Omeramsc: fixed

docs/guidelines/ai-assisted-development.md
```diff
@@ -0,0 +1,53 @@
+# AI-Assisted Development — Lessons Learned
```
rrosatti: hmm, I believe this file is not necessary. If we feel like we need to give a better guidance for AI assistants, then we could update the `docs/best-practices.md` file 🤔 (or other file(s))

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
```diff
@@ -0,0 +1,152 @@
...
+    });
+    compResult?.successes?.forEach((v) => {
+      acc.push({
+        title: v.metadata?.title,
+        description: v.metadata?.description,
+        status: CONFORMA_RESULT_STATUS.successes,
+        component: compResult.name,
+        collection: v.metadata?.collections,
+        image: compResult.containerImage,
+        code: v.metadata?.code,
```
rrosatti: it seems we're kinda duplicating this mapping 3 times, how about creating some utility function that would do the mapping? Something like this:

```tsx
const mapToUIConformaData = (v: ConformaRule, compResult: ComponentConformaResult, status: CONFORMA_RESULT_STATUS): UIConformaData => ({
  title: v.metadata?.title,
  description: v.metadata?.description,
  status,
  timestamp: v.metadata?.effective_on,
  component: compResult.name,
  msg: v.msg,
  collection: v.metadata?.collections,
  solution: v.metadata?.solution,
  image: compResult.containerImage,
  code: v.metadata?.code,
});
```

and use it like this:

```tsx
export const mapConformaResultData = (
  conformaResult: ComponentConformaResult[],
): UIConformaData[] => {
  return conformaResult.reduce((acc, compResult) => {
    compResult?.violations?.forEach((v) => {
      acc.push(mapToUIConformaData(v, compResult, CONFORMA_RESULT_STATUS.violations));
    });
    compResult?.warnings?.forEach((v) => {
      acc.push(mapToUIConformaData(v, compResult, CONFORMA_RESULT_STATUS.warnings));
    });
    compResult?.successes?.forEach((v) => {
      acc.push(mapToUIConformaData(v, compResult, CONFORMA_RESULT_STATUS.successes));
    });
    return acc;
  }, [] as UIConformaData[]);
};
```

I tested locally and it seems to work fine, but please check/test on your end as well :)

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
rrosatti: actually, I thought that we'd be able to replace the whole code in this try/catch block with the `fetchResourceWithK8sAndKubeArchive` utility - but it seems it won't work here as-is

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
rrosatti: were you able to check if the place that currently uses the `useConformaResult` hook keeps working fine?

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
```diff
@@ -0,0 +1,175 @@
...
+    );
+  }
+
+  const isEmpty = allResults.length === 0;
+
+  return (
+    <>
+      <PageSection variant={PageSectionVariants.light}>
+        <TextContent>
+          <Title headingLevel="h2" size="xl">
```
rrosatti: the title style is different from the other tabs, we should set the following props:

```tsx
headingLevel="h4" className="pf-v5-c-title pf-v5-u-mt-lg pf-v5-u-mb-lg" size="lg"
```

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
```diff
@@ -0,0 +1,175 @@
...
+        </Bullseye>
+      </PageSection>
+    );
+  }
+
+  const isEmpty = allResults.length === 0;
+
+  return (
+    <>
+      <PageSection variant={PageSectionVariants.light}>
```
rrosatti: we need to set `padding={{ default: 'noPadding' }}` here

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
```diff
@@ -0,0 +1,175 @@
...
+            totalComponents={totalComponents}
+            totalFailed={totalFailed}
+            totalViolations={totalViolations}
+            totalWarnings={totalWarnings}
+            totalSuccesses={totalSuccesses}
+          />
+        </div>
+      </PageSection>
+
+      <PageSection isFilled>
```
rrosatti: same here, we need to set `padding={{ default: 'noPadding' }}`

src/components/Conforma/useConformaResult.tsx
```diff
@@ -49,113 +47,41 @@ export const useConformaResultFromLogs = (
...
-          },
-        }
-      : null;
-  }, [podName, namespace]);
+  // Keep a ref to the latest taskRun so the effect body always reads the
+  // current value without adding the (potentially unstable) object reference
+  // to the dep array. A stable primitive derived from the task run is used as
+  // the dep instead: prefer uid when available, fall back to podName so that
+  // task runs without metadata (e.g. some test/edge-case runs) are still served.
+  const taskRunRef = React.useRef(taskRun);
```
rrosatti, line 59: did you face any issues with the previous implementation? 🤔

rrosatti (commented):
thanks for working on the changes, I left a few more comments! :)

docs/guidelines/ai-assisted-development.md
Omeramsc: Sure thing, I do have some possible rule\agents.md suggestions, but this is outside of the scope of this.

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
Omeramsc: fixed

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
Omeramsc: fixed

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
Omeramsc: fixed

src/components/Conforma/ConformaResultsTab/conforma-fetchers.ts
Omeramsc: Yes, verified with unit tests and with manual checks on staging. The Security tab rendered as expected.

src/components/Conforma/useConformaResult.tsx
Omeramsc: No functional issues. The identical logic already existed as `resolveConformaResultFromTaskRun()` in `conforma-fetchers.ts` for the application-level tab. I consolidated to a single call to avoid maintaining the same fallback chain in two places.

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
```diff
@@ -0,0 +1,175 @@
...
+    );
+  }
+
+  const isEmpty = allResults.length === 0;
+
+  return (
+    <>
+      <PageSection variant={PageSectionVariants.light} padding={{ default: 'noPadding' }}>
+        <TextContent>
+          <Title headingLevel="h4" className="pf-v5-c-title pf-v5-u-mt-lg pf-v5-u-mb-lg" size="lg">
```
rrosatti: my bad, actually `headingLevel` should be `h3` and use `pf-v5-u-mb-sm` instead of `pf-v5-u-mb-lg`

src/components/Conforma/ConformaResultsTab/ConformaGroupedTable.tsx
rrosatti: could you please create a ticket for doing this refactor? (use TableV2 once it's ready/merged) :)

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
```diff
@@ -0,0 +1,113 @@
...
+        selected={groupBy}
+        isOpen={isGroupByOpen}
+        onOpenChange={setIsGroupByOpen}
+      >
+        <SelectList>
+          <SelectOption value="rule">Rule</SelectOption>
+          <SelectOption value="component">Component</SelectOption>
+        </SelectList>
+      </Select>
+      <Button
```
rrosatti, line 97: nit: how about having only one button here that would change from "Expand/Collapse all"?

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
```diff
@@ -0,0 +1,113 @@
...
+        variant={ButtonVariant.link}
```
rrosatti: I believe using `variant={ButtonVariant.primary}` would be better here

rrosatti (commented):
I left a few minor comments regarding UI/UX. Overall the PR changes look good! :)

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
Omeramsc: done

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
Omeramsc: actually a very good suggestion :) fixed

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
Omeramsc: fixed

src/components/Conforma/ConformaResultsTab/ConformaGroupedTable.tsx
Omeramsc: sure thing, but it under my epic, and blocked by the actual feature https://redhat.atlassian.net/browse/KFLUXSE-276

src/components/Conforma/ConformaResultsTab/ConformaGroupedTable.tsx
```diff
@@ -0,0 +1,125 @@
...
+    </Thead>
+    <Tbody>
+      {rows.map((row, idx) => (
+        <Tr key={`${row.component}-${row.title}-${idx}`}>
+          <Td dataLabel="Rule">
+            <div>
+              <strong>{row.title ?? '-'}</strong>
+            </div>
+            {row.description && (
+              <div className="conforma-grouped-table__secondary-text">{row.description}</div>
```
rrosatti: maybe instead of creating this "custom text style",  how about using the `Text` component from `@patternfly/react-core` lib? 

`<Text component={TextVariants.p}>{row.description}</Text>` should work fine here :)

src/components/Conforma/ConformaResultsTab/ConformaGroupedTable.tsx
```diff
@@ -0,0 +1,125 @@
...
+          </Td>
+          <Td dataLabel="Component">{row.component}</Td>
+          <Td dataLabel="Image">
+            {row.image ? <Truncate content={row.image} /> : '-'}
+          </Td>
+          <Td dataLabel="Status">{getRuleStatus(row.status)}</Td>
+          <Td dataLabel="Message">
+            <div>{row.msg ?? '-'}</div>
+            {row.solution && (
+              <div className="conforma-grouped-table__secondary-text">
```
rrosatti: same here, so we can delete the css style

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.scss
```diff
@@ -0,0 +1,12 @@
+.conforma-results-tab {
+  &__summary-wrapper {
+    margin-top: var(--pf-v5-global--spacer--md);
+  }
+}
+
+.conforma-grouped-table {
+  &__secondary-text {
```
rrosatti: as mentioned in the previous comments, we might be able to use the `Text` component from PF lib. So this css style can be removed

rrosatti (dismissed):
I left some comments, but it should be fine if you do it in a separate PR. But if you decide to do in this PR, just let me know and I'll re-review/approve again. Thanks for the work on the PR, it's a cool feature :)

src/components/Conforma/ConformaResultsTab/ConformaGroupedTable.tsx
Omeramsc: Fixed :)

src/components/Conforma/ConformaResultsTab/ConformaGroupedTable.tsx
Omeramsc: Fixed :)

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.scss
Omeramsc: Fixed :)

rrosatti (approved):

sahil143 (approved):

JoaoPedroPP (approved):
LGTM

PR #868, opened 2026-05-07
feat(KFLUXUI-708): feat(KFLUXUI-1222):Support for multi-subject RoleBindings, change access logic

milantaky:
## Fixes 

https://redhat.atlassian.net/browse/KFLUXUI-708
https://redhat.atlassian.net/browse/KFLUXUI-1222

## Description

This PR improves the User Access experience to better reflect how RBAC is represented in Kubernetes and to support bulk `Change access` actions.

The table structure was intentionally changed compared to most other Konflux tables: instead of rendering 1 row per RoleBinding, the view now renders 1 row per subject (expanding a single RoleBinding with multiple users into multiple rows). This is different from typical Konflux tables where each row maps cleanly to a single resource. Here, that pattern breaks down because a RoleBinding can legitimately include multiple subjects, and the user-facing operation (“change access for these users”) is subject-centric. The expanded row model enables:

- Correctly showing all users on shared RoleBindings (fixes incomplete visibility).
- `Row selection` that matches the action semantics (select users, not just bindings).
- `Filtering` by username / role binding name while keeping selection behavior predictable.

`Change access` is implemented as a `delete-and-recreate` workflow because Kubernetes RoleBindings don’t allow changing roleRef in-place. On save, the logic:

- Collects all RoleBindings affected by the selected users.
- Deletes the affected RoleBindings.
- Creates new RoleBinding(s) for the selected users with the chosen role.
- If a deleted RoleBinding originally contained additional unselected users, it recreates RoleBinding(s) for those users to preserve their previous access.
- Prevents unsupported downgrades for multi-select (while still allowing downgrade when exactly one user is selected), and avoids recreating bindings for users who already have the target role.

## Type of change

- [X] Feature
- [ ] Bugfix
- [ ] Build related changes

`UX approved`
All subjects displayed from `konflux-contributor-many-men2-user-actions`:
Screenshot 2026-05-07 at 11 43 04

Change access modal:
Screenshot 2026-05-07 at 11 45 28

## How to test or reproduce?

- In page `User Access` find a RoleBinding with multiple subjects
    - If you don't have one, add one through CLI or 
    - In file `UserAccessListView.tsx` from line 104
    
    ```
    // const [roleBindings, loaded, error] = useRoleBindings(namespace);
   const roleBindings = mockRoleBindingsWithMultipleUsers;
   const loaded = true;
   const error = null;
  ```
- See that all subjects are displayed

Change access:
- Create test users eg. user-a, user-b (it is necessary to use the CLI for RoleBindings with multiple subjects, since there is no way to do it from UI)
- Select one or more users and change their access
- Check

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## Summary by CodeRabbit

* **New Features**
  * Table shows one row per subject, selectable rows, dual-mode filtering (username or role-binding name), and a bulk "Change access" modal with per-row summaries and controlled save behavior.

* **Improvements**
  * Enforced role-precedence rules, safer change-role workflow with create-before-delete and rollback on failure, preserves other users in shared bindings, and automatic pruning of selections when filters change.

* **Style**
  * Layout spacing tweaks for modal user list and toolbar.

* **Tests**
  * Expanded unit and integration coverage for list view, modal flows, filtering, selection, and rollback scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/konflux-ci/konflux-ui/pull/868?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/components/UserAccess/UserAccessListView.tsx
```diff
@@ -64,26 +91,155 @@ export const UserAccessListView: React.FC<React.PropsWithChildren<unknown>> = ()
...
+      const ok = !(rb.subjects?.length === 1 && rb.roleRef?.name === newRoleRef);
+      const username = rb.subjects?.[0]?.name;
+
+      // Has target role, will not be recreated, only rest of RBs deleted
+      if (!ok) {
+        uniqueSelectedUsernames.delete(username);
+      }
+
+      return ok;
+    });
```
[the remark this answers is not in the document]
milantaky: This filters users who already have a single role binding with the target role, so it it okay.

src/components/UserAccess/UserAccessForm/form-utils.ts
```diff
@@ -100,12 +100,39 @@ export const deleteRB = async (roleBinding: RoleBinding, dryRun?: boolean): Prom
...
+  };
+
+  return K8sQueryCreateResource({
+    model: RoleBindingModel,
+    queryOptions: {
+      ns: md.namespace ?? '',
+    },
+    resource,
+  });
+};
```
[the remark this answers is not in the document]
milantaky: @coderabbitai I updated the namespace line and didn't add the checks because both are present here, resolving.

src/components/UserAccess/UserAccessListView.tsx
```diff
@@ -92,51 +224,146 @@ export const UserAccessListView: React.FC<React.PropsWithChildren<unknown>> = ()
...
           aria-label="User access list"
-          Header={RBListHeader}
-          Row={RBListRow}
-          loaded
-          getRowProps={(obj: RoleBinding) => ({
-            id: obj.metadata.name,
-          })}
-        />
+          variant="compact"
+          gridBreakPoint={TableGridBreakpoint.none}
```
sahil143, line 359: **Critical:** This PR bypasses the shared `Table` component and inlines PatternFly primitives (`<Table>`, `<Thead>`, `<Tbody>`, `<Tr>`, `<Td>`) with hand-rolled selection state (`Set<string>`, `onSelectAll`, `onSelectRow`, filter-prune effect).

The data transform (`expandRoleBindingsToTableRows`) is the right approach — converting one-row-per-binding to one-row-per-subject is purely a data concern and it's cleanly separated. That should stay.

But why wasn't the shared `Table` component extended with selection support instead of creating a new inline table? Selection (checkbox column, select-all, prune-on-filter) is a generic table capability, not specific to user access. Inlining it here means:

- No consistent table visuals across the app — every future table with selection or expand/collapse will be a new bespoke component
- A virtualized table with selection or expand/collapse becomes a separate effort instead of composing on one foundation
- The selection state management gets copy-pasted the next time someone needs it

The shared `Table` component should be extended to support row selection as a composable feature. Feed the transformed data into that, keeping consistent rendering behavior app-wide.

src/components/UserAccess/UserAccessChangeRoleModal.tsx
```diff
@@ -0,0 +1,298 @@
...
+export const splitRowKey = (
+  rowKey: string,
+): {
+  roleRefName: string;
+  roleName: string;
+  index: string;
+  subjectKind: string;
+  username: string;
+} => {
+  const segments = rowKey.split('__');
```
sahil143: **Suggestion:** `splitRowKey` relies on `rowKey.split('__')` with positional indexing. If any segment (username, binding name) ever contains `__`, this silently corrupts the parsed values.

K8s DNS-compatible names can't have `__`, but this is an implicit contract with no validation at the point where row keys are constructed or parsed. Consider either:
- Documenting this invariant explicitly at both construction (`expandRoleBindingsToTableRows`) and parsing sites
- Storing structured data (e.g., passing the `UserAccessTableRow` object to the modal) instead of encoding/decoding a delimiter-separated string

src/utils/rbac.ts
```diff
@@ -195,3 +195,9 @@ export const createLoaderWithAccessCheck =
     }
     return defer({ accessCheck: allowed, data: loader(args) });
   };
+
+export const ROLE_WEIGHT_MAP: Record<string, number> = {
```
sahil143: **Suggestion:** `ROLE_WEIGHT_MAP` is hardcoded. If a new role is added to the cluster (e.g., `Viewer`), it won't appear here and `ROLE_WEIGHT_MAP[displayName]` returns `undefined`.

The downgrade check treats unknown roles inconsistently — in `highestAffectedRoleWeight` they're filtered out via `filter((weight): weight is number => weight !== undefined)`, but elsewhere `ROLE_WEIGHT_MAP[role] ?? 0` treats them as the lowest tier. A user with an unknown role could bypass the downgrade guard entirely.

Consider deriving weights from the role data source, or at minimum treating unknown roles as an error state in the modal rather than silently assigning them weight 0.

src/hooks/useChangeAccessSave.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function useChangeAccessSave({
+  namespace,
+  currentRoleMap,
+  roleBindings,
+  roleMap,
+  selectedRowKeys,
+  setSelectedRowKeys,
+}: UseChangeAccessSaveParams): (newRoleRef: string) => Promise<void> {
+  return React.useCallback(
```
sahil143, line 28: **Suggestion:** This `useCallback` closes over `selectedRowKeys` and `roleBindings` at render time. The returned function is passed as `onSave` to the modal and invoked after user interaction (select role → click save). If bindings change between modal open and save click (another tab modifies access, polling refreshes data), the callback operates on stale state.

Consider snapshotting `selectedRowKeys` and `roleBindings` at modal-open time, or at minimum documenting this assumption.

src/components/UserAccess/userAccessChangeAccessSave.ts
```diff
@@ -0,0 +1,167 @@
...
+  return bindings.filter((rb) => rb.subjects?.some((subject) => users.has(subject.name)));
+}
+
+export function roleBindingHasNonUserSubject(rb: RoleBinding): boolean {
+  return (rb.subjects ?? []).some((subject) => subject.kind !== 'User');
+}
+
+/** Deep clone for rollback (Jest/jsdom may not expose `structuredClone`). */
+export function cloneRoleBindingSnapshot(rb: RoleBinding): RoleBinding {
+  return JSON.parse(JSON.stringify(rb));
```
sahil143: **Nitpick:** `JSON.parse(JSON.stringify(rb))` for cloning, but `restoreRB` in `form-utils.ts` already reconstructs a clean copy by explicitly picking fields (`name`, `namespace`, `labels`, `annotations`, `roleRef`, `subjects`). The deep clone is only needed to prevent mutation of the original reference between snapshot and restore — a shallow spread of `subjects` and `roleRef` would suffice since the K8s API calls don't mutate the input object.

src/components/UserAccess/userAccessChangeAccessSave.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    for (const rb of toDelete) {
+      await deleteRB(rb);
+      deletedSnapshots.push(cloneRoleBindingSnapshot(rb));
+    }
+
+    onSuccessClearSelection();
+  } catch (err: unknown) {
+    await restoreBindingsBestEffort(deletedSnapshots, namespace);
+    await deleteBindingsBestEffort(
```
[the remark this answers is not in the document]
milantaky: @coderabbitai Resolving — no change. Create-before-delete is deliberate (no access gap; rollback on failure). hasSingleTargetRole / usernamesSkippingTargetRoleCreate, userHasRoleBindingOutsideDeletions, and the non-User-subject guard prevent the AlreadyExists split case you describe. Tests cover ordering and rollback.

src/components/UserAccess/userAccessChangeAccessSave.ts
```diff
@@ -0,0 +1,194 @@
...
+
+export function getUniqueSelectedUsers(rowKeys: Set<string>): Set<string> {
+  return new Set([...rowKeys].map((rowKey) => splitRowKey(rowKey).username));
+}
+
+export function getAllAffectedRoleBindings(
+  users: Set<string>,
+  bindings: RoleBinding[],
+): RoleBinding[] {
+  return bindings.filter((rb) => rb.subjects?.some((subject) => users.has(subject.name)));
```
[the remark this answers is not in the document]
milantaky: Resolving. Change access is User-only. With User/userc (Contributor) selected and a separate Group/userc (Maintainer) binding in the namespace, save includes that Group binding by name and correctly blocks with the non-User-subject error: Fix/split the Group binding in the cluster, then retry the User change. Kind+name matching is out of scope for this PR.

src/components/UserAccess/RBListHeader.tsx
JoaoPedroPP, line 1: I understand that you intentionally refactor the tables to do not use the standard/pattern we currently have but what is the reason for the refactor? Why do we need to use this way instead of adjusting the current Table components we have?

milantaky: Hey @JoaoPedroPP! The changes would be too much for this PR (poor task decomposition:( ), but I talked to @sahil143, and right after this PR is merged, I have a ticket ready to address this, and extend our shared table component to support row selection (https://redhat.atlassian.net/browse/KFLUXUI-1346)

StanislavJochman (dismissed):
LGTM

StanislavJochman (dismissed):

sahil143 (changes requested):
Requesting changes.

src/components/UserAccess/UserAccessChangeRoleModal.tsx
```diff
@@ -0,0 +1,341 @@
...
+                      isDisabled={roleSelectOptions.length === 0 || isSaving}
+                      isFullWidth
+                    >
+                      {modalSelectedRoleRef
+                        ? roleMap.roleMap[modalSelectedRoleRef] ?? modalSelectedRoleRef
+                        : 'Select a role'}
+                    </MenuToggle>
+                  )}
+                  onSelect={(_, val) => {
+                    setModalSelectedRoleRef(val as string);
```
sahil143: setModalSelectedRoleRef(val as string);
PF Select's onSelect val is string | number | undefined. Use a runtime guard:
if (typeof val === 'string') { setModalSelectedRoleRef(val); }

src/components/UserAccess/userAccessTableRows.ts
```diff
@@ -0,0 +1,148 @@
...
+  roleBindingName?: string;
+};
+
+export function filterUserAccessRows(
+  rows: UserAccessTableRow[],
+  filters: UserAccessRowFilters,
+): UserAccessTableRow[] {
+  const usernameQ = filters.username?.trim().toLowerCase() ?? '';
+  if (usernameQ) {
+    return rows.filter((row) => row.subject && row.subject.name.toLowerCase().includes(usernameQ));
```
sahil143: should use textMatch util to filter based on user name

src/components/UserAccess/userAccessTableRows.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  const roleBindingQ = filters.roleBindingName?.trim().toLowerCase() ?? '';
+  if (roleBindingQ) {
+    return rows.filter((row) => {
+      const name = row.roleBinding.metadata?.name?.toLowerCase() ?? '';
+      return name.includes(roleBindingQ);
```
sahil143: same as above

src/components/UserAccess/UserAccessChangeRoleModal.tsx
```diff
@@ -0,0 +1,341 @@
...
+      getRoleRefWeight(modalSelectedRoleRef, roleRefWeights) === undefined
+    ) {
+      refs.add(modalSelectedRoleRef);
+    }
+    return [...refs];
+  }, [allAffectedRoleBindings, modalSelectedRoleRef, roleRefWeights]);
+
+  const hasUnrankedRoles = unrankedRoleRefs.length > 0;
+
+  React.useEffect(() => {
```
sahil143: Why is this useEffect needed?

src/components/UserAccess/UserAccessChangeRoleModal.tsx
milantaky: updated

src/components/UserAccess/userAccessTableRows.ts
milantaky: updated

src/components/UserAccess/userAccessTableRows.ts
milantaky: updated

src/components/UserAccess/UserAccessChangeRoleModal.tsx
milantaky: It is not actually, thank you for noticing. Removed

sahil143 (approved):

StanislavJochman (approved):

PR #849, opened 2026-04-30
feat(KFLUXUI-1277): refactor log data flow to use structured per-container data

rrosatti:
## Fixes
Fixes: https://issues.redhat.com/browse/KFLUXUI-1277

## Description

Replace the flat string log pipeline with a structured `LogSection[]` type that preserves per-container structure through the component tree.

**Before:** `processLogs()` in `Logs.tsx` split each container's logs by `\n`, added string indentation (`'  ' + line`), and joined everything into a single string — which `LogViewer` and `VirtualizedLogContent` then split again (3 redundant split operations per render cycle).

**After:** `Logs.tsx` builds a `LogSection[]` via `useMemo` (one entry per container with raw `data: string`). `VirtualizedLogContent` handles all processing in a single pass (normalize line endings, strip ANSI codes, split into lines), and uses CSS `padding-left` for indentation instead of string manipulation.

- Add `LogSection` type (`{ containerName: string; data: string }`)
- Remove `processLogs()`, its debounce wrapper, and related state (`formattedLogs`, `processingLogs`, `containersRef`)
- `LogViewer` accepts `sections: LogSection[]` as the single data API
- `TektonTaskRunLog` wraps its pre-aggregated string as a single-element `LogSection[]`
- ANSI stripping and line-ending normalization applied once in the render layer
- `VirtualizedLogContent` builds a flat line list from sections, tracking container headers via index set for CSS indentation

## Type of change

- [x] Refactoring (no functional changes, no api changes)

### KubeArchive (sections path) — rendered logs, search, download

https://github.com/user-attachments/assets/df5ba026-caa9-42e5-aba5-e5ea3d72d5d1

### Tekton Results (data string path) — rendered logs, search

https://github.com/user-attachments/assets/f5a02827-67dc-4ab3-99d2-c206c715208c

### Live log streaming — WebSocket fetch, auto-scroll

https://github.com/user-attachments/assets/5f0e2e2c-e08b-4390-ac09-779fc3719c30

## How to test or reproduce?

- [ ] Open a pipeline run with multiple task steps — logs should display per-container with container names visible
- [ ] Search for text spanning multiple containers — matches should highlight correctly
- [ ] Search for a container name — should match and highlight
- [ ] Verify auto-scroll works during streaming (running pipeline)
- [ ] Download logs — file should contain all container logs
- [ ] Open a Tekton Results log (completed pipeline via Tekton Results API) — should still render correctly

## Browser conformance:
- [x] Chrome
- [x] Firefox
[...]

PR #895, opened 2026-05-19
feat(KFLUXUI-1176): fold/unfold logs step

janaki29:
## Fixes 
[KFLUXUI-1176](https://redhat.atlassian.net/browse/KFLUXUI-1176)
[KFLUXUI-1337](https://redhat.atlassian.net/browse/KFLUXUI-1337)
[KFLUXUI-1338](https://redhat.atlassian.net/browse/KFLUXUI-1338)

## Description

1. Replaces the flat single-stream log viewer for pipeline run logs with a VS Code-style foldable step view.
2. Each container step in a pipeline run is now a collapsible section with a fold/unfold icon (▼/▶) next to the line number
3. Line numbers are continuous across all steps — section headers consume a line number, content lines follow sequentially
4. Scrolling through a step's lines pins the step header at the top (sticky scroll with push-up animation when the next step approaches)
5. All steps start expanded; new steps auto-expand as they come online during a live run
6. Collapsed steps show ··· N lines hidden so you know what was omitted
7. Syntax highlighting (Prism tokeniser) and the dark/light theme toggle both work identically to the existing viewer

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

https://github.com/user-attachments/assets/28029648-60b3-488c-8cc4-f6624b1c0e73

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

[KFLUXUI-1176]: https://redhat.atlassian.net/browse/KFLUXUI-1176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KFLUXUI-1337]: https://redhat.atlassian.net/browse/KFLUXUI-1337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KFLUXUI-1338]: https://redhat.atlassian.net/browse/KFLUXUI-1338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Foldable pipeline log viewer: collapsible sections, sticky header with push-up transition, auto-expand, scroll-to-row, virtualized rendering, and preserved search/highlighting across sections.

* **Style**
  * New stylesheet for foldable log layout, theming, hover/focus states, gutters, line visuals and collapsed-state indicators.

* **Refactor**
  * Log ingestion now produces sectioned log data consumed by the viewer (supports both structured sections and single-stream mode).

* **Tests**
  * Updated tests to validate sectioned log handling and related behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/konflux-ci/konflux-ui/pull/895?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/shared/components/pipeline-run-logs/logs/LogViewer.tsx
```diff
@@ -46,9 +47,12 @@ import './LogViewer.scss';
 // eslint-disable-next-line no-control-regex
 const ANSI_ESCAPE_REGEX = /\u001b\[[0-9;]*m/g;
 
+export type LogSection = { readonly name: string; readonly data: string };
+
 export type Props = {
   showSearch?: boolean;
-  data: string;
+  data?: string;
+  logSections?: LogSection[];
```
sahil143: DRY! Why is same data being sent in two different formats?

sahil143 (changes requested):
What's the benefit of creating another virtualized log viewer? On top of that PR is introducing multiple regressions (missing keyboard navigation, missing url navigation, dead code, two parallel rendering paths). Adding Foldable as a parallel component causes issues when we need to fix something, two components will require fixes for one bug.

src/shared/components/pipeline-run-logs/logs/FoldableLogView.tsx
```diff
@@ -0,0 +1,429 @@
+import React from 'react';
```
sahil143: **Critical: This component reimplements VirtualizedLogViewer internals instead of composing with them.**

The existing `virtualized-log-viewer/` directory is well-decomposed into focused hooks — `useKeyboardNavigation`, `useLineNumberNavigation`, `LineNumberGutter`, `virtualization-utils`, etc. This component imports only 4 of them and reimplements the rest in a single 429-line file.

What's lost compared to `VirtualizedLogContent`:
- **Keyboard navigation** (`useKeyboardNavigation`) — no PageUp/Down, Home/End, arrow keys
- **Line number URL navigation** (`useLineNumberNavigation`) — no `#L123` linking, no shift-click ranges
- **Clickable line numbers** (`LineNumberGutter`) — replaced with a plain `<span>` (`NumberCol`) that has no click handler or highlight state
- **Dynamic line height estimation** (`virtualization-utils`) — hardcoded `ROW_HEIGHT = 20`

This creates two parallel rendering paths that will drift. Any future fix to keyboard nav, gutter behavior, or line linking would need to be applied in both places.

The "foldable" concept is orthogonal to log rendering. Consider composing instead:
1. A **data transform** that takes `LogSection[]` + fold state and produces a flat row list with section-header rows injected
2. The **existing `VirtualizedLogContent`** to render those rows (it already handles virtualization, scroll, keyboard, search, line numbers)
3. A **sticky-header overlay** as a separate concern driven by scroll position

Each piece would be independently testable, and the fold feature wouldn't regress existing functionality.

src/shared/components/pipeline-run-logs/logs/FoldableLogView.tsx
```diff
@@ -0,0 +1,429 @@
...
+
+  const { stickyRow, pushUpOffset } = React.useMemo(() => {
+    if (scrollTop <= 0 || sectionHeaderRowIndices.length === 0) {
+      return { stickyRow: null, pushUpOffset: 0 };
+    }
+
+    let currentBucket = -1;
+    for (let j = 0; j < sectionHeaderRowIndices.length; j++) {
+      const headerTop = sectionHeaderRowIndices[j] * ROW_HEIGHT;
+      if (headerTop < scrollTop) {
```
sahil143: **Bug: Sticky header position assumes fixed row heights but virtualizer uses dynamic measurement.**

```tsx
const headerTop = sectionHeaderRowIndices[j] * ROW_HEIGHT;
```

This assumes every row is exactly 20px. But the virtualizer uses `measureElement` for dynamic sizing, and content rows have `word-break: break-all` + `white-space: pre-wrap` — so any long line that wraps will be taller than 20px. Once that happens, every sticky header position below the wrapped line will be wrong, causing the header to appear/disappear at incorrect scroll offsets.

src/shared/components/pipeline-run-logs/logs/FoldableLogView.tsx
```diff
@@ -0,0 +1,429 @@
...
+    });
+    return map;
+  }, [displayRows]);
+
+  // If a search result lands inside a collapsed section, expand it so the row
+  // becomes visible before useVirtualizedScroll tries to scroll to it.
+  // Use a ref to avoid listing expandedSections in deps (prevents re-trigger loop).
+  const expandedSectionsRef = React.useRef(expandedSections);
+  expandedSectionsRef.current = expandedSections;
+  React.useEffect(() => {
```
sahil143: **Bug: This effect can re-expand a section the user just collapsed.**

`rowIndexToFlatLineIndex` and `flatIndexToSectionIndex` are rebuilt whenever `logSections` changes (new logs arriving). This re-triggers the effect with the current `scrollToRow`, which may still point at the section the user just manually collapsed — re-expanding it against the user's intent.

src/shared/components/pipeline-run-logs/logs/FoldableLogView.tsx
```diff
@@ -0,0 +1,429 @@
...
+  readonly height: number;
+  readonly scrollToRow?: number;
+  readonly onScroll?: (props: {
+    scrollDirection: 'forward' | 'backward';
+    scrollOffset: number;
+    scrollUpdateWasRequested: boolean;
+  }) => void;
+}
+
+export const FoldableLogView: React.FC<FoldableLogViewProps> = ({
```
sahil143: **Missing: No unit tests for FoldableLogView.**

This is 429 lines of non-trivial logic — expand/collapse state, sticky header math, three mapping data structures, scroll-to-row translation, auto-expand on search. The test changes in `Logs.spec.tsx` mock out `LogViewer` entirely, so none of this is covered.

src/shared/components/pipeline-run-logs/logs/FoldableLogView.tsx
```diff
@@ -0,0 +1,429 @@
...
+import { useLineRenderer } from '~/shared/components/virtualized-log-viewer/useLineRenderer';
+import { useSearchRegex } from '~/shared/components/virtualized-log-viewer/useSearchRegex';
+import { useTokenization } from '~/shared/components/virtualized-log-viewer/useTokenization';
+import { useVirtualizedScroll } from '~/shared/components/virtualized-log-viewer/useVirtualizedScroll';
+import type { LogSection } from './LogViewer';
+
+import './FoldableLogView.scss';
+
+// eslint-disable-next-line no-control-regex
+const ANSI_ESCAPE_REGEX = /\u001b\[[0-9;]*m/g;
```
sahil143: Nitpick: This regex is already defined in `LogViewer.tsx:48`. Extract to a shared constant.

src/shared/components/pipeline-run-logs/logs/FoldableLogView.scss
```diff
@@ -0,0 +1,158 @@
...
+
+// Sticky header overlay — absolutely positioned on the wrapper so it never
+// scrolls, always visible at the top while you're inside a step.
+.foldable-log__sticky-overlay {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  z-index: 10;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.4);
```
sahil143: Suggestion: `rgba(0, 0, 0, 0.4)` here and `rgba(128, 128, 128, 0.1)` on the hover state (line 128) bypass PatternFly tokens. The rest of this file correctly uses PF CSS variables — these should too, for theme consistency.

src/shared/components/pipeline-run-logs/logs/Logs.tsx
```diff
@@ -13,12 +13,13 @@ import { PodModel } from '../../../../models/pod';
...
 import { PodKind, ContainerSpec, ContainerStatus } from '../../types';
 import { containerToLogSourceStatus, LOG_SOURCE_TERMINATED } from '../utils';
-import LogViewer, { type Props as LogViewerProps } from './LogViewer';
+import LogViewer, { type LogSection, type Props as LogViewerProps } from './LogViewer';
 
 type LogSources = { [containerName: string]: string };
 
 const WEB_SOCKET_RETRY_COUNT = 5;
 
+// Kept exported for backward compat with existing tests
```
sahil143: Suggestion: `processLogs` is no longer called in production code. Keeping dead production code for test convenience is the wrong trade-off — either remove it and update the tests, or move it to a test utility file.

src/shared/components/pipeline-run-logs/logs/FoldableLogView.tsx
```diff
@@ -0,0 +1,429 @@
...
+  //   lines[1..n0]       = section 0 content
+  //   lines[n0+1]        = '' (separator, omitted after last section)
+  //   lines[n0+2]        = section 1 name
+  //   lines[n0+3..n0+n1+2] = section 1 content
+  //   ...
+  //
+  // rowIndexToFlatLineIndex maps a rowIndex (in the combined-string space) to
+  // the matching flatLineIndex in allLines (content only, no names/separators).
+
+  const rowIndexToFlatLineIndex = React.useMemo(() => {
```
sahil143: Suggestion: `rowIndexToFlatLineIndex`, `flatIndexToSectionIndex` (line 220), and `flatIndexToDisplayRowIndex` (line 270) each iterate over all sections/lines independently. These could be built in a single pass, which also makes the index relationships easier to reason about.

sahil143 (commented):
Is this PR dependent on https://github.com/konflux-ci/konflux-ui/pull/849

janaki29:
@sahil143 No, this PR is not dependent on #849. Both PRs independently reached the same structural conclusion — replacing the flat string pipeline with a LogSection[] type — but our implementation is self-contained and builds on top of current main without requiring #849 to be present.

There are conflicts rather than a dependency: both PRs touch LogViewer.tsx, VirtualizedLogContent.tsx, Logs.tsx, and types.ts. If #849 merges first, we'll rebase and resolve those conflicts, but our feature (foldable steps, sticky headers, auto-collapse on completion) works either way.

Our implementation is actually a superset of what #849 proposes — it covers the same structured data pipeline and additionally adds the VS Code-style fold/unfold UI. So we don't need to adopt anything from #849; we just need to rebase if it lands first.

JoaoPedroPP (commented):
Nice PR and this will certainly help user to move though logs. I have a question, is our intention to set as default behavior collapse the lines when some user looking for an old log or when one step just finished like the image below?

Image

Personally I prefer to keep streaming logs and let users decide if they want to collapse lines when pipelinerun still running but I am fine with this implementation. Does Pankaj have an opinion on this topic?

marcin-michal:
> Nice PR and this will certainly help user to move though logs. I have a question, is our intention to set as default behavior collapse the lines when some user looking for an old log or when one step just finished like the image below?
> Image
> 
> Personally I prefer to keep streaming logs and let users decide if they want to collapse lines when pipelinerun still running but I am fine with this implementation. Does Pankaj have an opinion on this topic?

@JoaoPedroPP the default behaviour for logs in Github is the same (finished tasks are folded by default), so I think it's fine here as well
image

PR #994, opened 2026-06-08
feat(KFLUXUI-1176): foldable log steps

janaki29:
## Feature
[KFLUXUI-1176](https://redhat.atlassian.net/browse/KFLUXUI-1176)
[KFLUXUI-1338](https://redhat.atlassian.net/browse/KFLUXUI-1338)
[KFLUXUI-1337](https://redhat.atlassian.net/browse/KFLUXUI-1337)

## Description
Extends the virtualized log viewer with VS Code–style foldable pipeline steps, so multi-container task logs can be collapsed/expanded per step while keeping virtualization, search, line numbers, and sticky section headers.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

https://github.com/user-attachments/assets/5a20cd25-9980-4e45-ac9e-454804a5b2b3

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

[KFLUXUI-1176]: https://redhat.atlassian.net/browse/KFLUXUI-1176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KFLUXUI-1338]: https://redhat.atlassian.net/browse/KFLUXUI-1338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KFLUXUI-1337]: https://redhat.atlassian.net/browse/KFLUXUI-1337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Sectioned log view with per-section headers, downloadable logs that include headers, improved search across sections, and auto-expansion of sections when jumping to search matches.
  * Optional mapped line numbers and sticky section headers for clearer navigation and context.

* **Refactor**
  * Virtualization rewritten to render section headers, fold indicators and content rows; folding preserves completed sections and improves scroll/scroll-to-hash behavior.
  * Log text normalization for consistent display and download.

* **Tests**
  * Extensive test updates covering sectioned viewing, folding, search, download and virtualization behaviors.


src/shared/components/virtualized-log-viewer/section-log-ui.tsx
```diff
@@ -0,0 +1,91 @@
...
+      </FlexItem>
+    </Flex>
+  </Button>
+);
+
+export const FoldIndicatorLine: React.FC<{ lineCount: number }> = ({ lineCount }) => (
+  <Text
+    component="small"
+    className="pf-v5-c-log-viewer__text"
+    style={{ fontStyle: 'italic', color: 'var(--pf-v5-global--disabled-color--100)' }}
```
rrosatti: nit: I'd define these styles in a css class and use it together with `className`

src/shared/components/virtualized-log-viewer/SectionLogUI.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const StickySectionHeaderBar: React.FC<{
+  row: SectionHeaderRow;
+  pushUpOffset: number;
+  itemSize: number;
+  onToggle: () => void;
+}> = ({ row, pushUpOffset, itemSize, onToggle }) => (
+  <Flex
+    alignItems={{ default: 'alignItemsStretch' }}
+    gap={{ default: 'gapNone' }}
+    style={{
```
rrosatti, line 46: nit: can we use css class here?

src/shared/components/virtualized-log-viewer/SectionedVirtualRow.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onToggleSection,
+  renderLogLine,
+}) => {
+  const rowClassName = `pf-v5-c-log-viewer__list-item${
+    row.kind === 'content' && isLineHighlighted(row.globalLineNumber)
+      ? ' log-content__line--highlighted'
+      : ''
+  }`;
+
+  const rowProps = {
```
rrosatti, line 75: I'd memoize this object

src/shared/components/virtualized-log-viewer/sticky-section-header.ts
```diff
@@ -0,0 +1,56 @@
...
+}): { stickyRow: SectionHeaderRow | null; pushUpOffset: number } {
+  const { enabled, scrollTop, displayRows, virtualItems, itemSize } = options;
+
+  if (!enabled || scrollTop <= 0) {
+    return { stickyRow: null, pushUpOffset: 0 };
+  }
+
+  const sectionHeaderRowIndices: number[] = [];
+  for (let i = 0; i < displayRows.length; i++) {
+    if (displayRows[i].kind === 'section-header') sectionHeaderRowIndices.push(i);
```
rrosatti: nit: can we use the `isSectionHeaderRow` utility here? :)

```tsx
if (isSectionHeaderRow(displayRows[i])) { ...
```

src/shared/components/virtualized-log-viewer/sticky-section-header.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+  if (sectionHeaderRowIndices.length === 0) {
+    return { stickyRow: null, pushUpOffset: 0 };
+  }
+
+  const headerTopByIndex = new Map<number, number>();
+  for (const vItem of virtualItems) {
+    if (displayRows[vItem.index]?.kind === 'section-header') {
```
rrosatti: nit: same here 

```tsx
if (isSectionHeaderRow(displayRows[vItem.index])) { ...
```

src/shared/components/virtualized-log-viewer/VirtualizedLogContent.tsx
```diff
@@ -39,201 +47,183 @@ export const VirtualizedLogContent: React.FC<VirtualizedLogContentProps> = ({
...
+      if (row?.kind === 'content') return row.globalLineNumber;
+      return null;
+    },
+    [displayRows],
+  );
 
   return (
-    <>
-      {/* Hidden element to measure actual line height */}
+    <Flex style={{ position: 'relative' }}>
```
rrosatti: nit: I'd define it in a css class

rrosatti (commented):
that's a really cool feature! I left a couple of minor comments :)

also, I noticed that the line number navigation is not working:

https://github.com/user-attachments/assets/0baf7520-ca4f-43ff-8c9f-34ac7a3e0b41

we need to fix this.

src/shared/components/virtualized-log-viewer/useSectionFold.ts
```diff
@@ -0,0 +1,105 @@
...
+          newlyCompleted.forEach((i) => next.delete(i));
+        }
+      }
+
+      return next ?? prev;
+    });
+
+    prevSectionCountRef.current = currentSections.length;
+    prevCompletionRef.current = currentSections.map((s) => s.isCompleted ?? false);
+  }, [hasSections, sections]);
```
rrosatti, line 75: do we really need to have `sections` in the dep array? we're not using it in this `useEffect` 🤔

src/shared/components/virtualized-log-viewer/section-log-ui.tsx
```diff
@@ -0,0 +1,91 @@
...
+  >
+    ··· {lineCount} lines hidden
```
rrosatti: if there's 1 line, it'll show as "1 lines hidden". We could handle it here to be:

```tsx
··· {lineCount} {lineCount === 1 ? 'line' : 'lines'} hidden.
```

src/shared/components/pipeline-run-logs/logs/LogViewer.tsx
```diff
@@ -109,12 +112,6 @@ const LogViewer: React.FC<Props> = ({
   const [downloadAllStatus, setDownloadAllStatus] = React.useState(false);
   const [showShortcutHint, setShowShortcutHint] = React.useState(false);
 
-  const downloadData = React.useMemo(() => {
-    return sections
-      .map((s) => (s.containerName ? `${s.containerName}\n${s.data}` : s.data))
-      .join('\n\n');
-  }, [sections]);
-
```
sahil143, line 116: This seems to be correct and was fixed in Rodrigo's PR. Please revert this change.

sahil143 (changes requested):

src/shared/components/pipeline-run-logs/logs/LogViewer.tsx
```diff
@@ -89,16 +89,19 @@ const LogViewer: React.FC<Props> = ({
...
   const themeCheckboxId = React.useId();
 
-  // Auto-scroll and resume button logic
+  const { processedData, downloadData } = React.useMemo(
+    () => prepareLogViewerContent(sections),
+    [sections],
+  );
+
+  const lines = React.useMemo(() => processedData.split('\n'), [processedData]);
+
```
sahil143: prepareLogViewerContent is joining the logs with `\n` and this useMemo is again spliting the data, It's redundant work

src/shared/components/virtualized-log-viewer/SectionLogUI.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
```
milantaky, line 1: nit: file doesn't use PascalCase

milantaky (commented):
Left a small nit, checkout Rodrigo's and Sahil's comments, otherwise LGTM, working as expected:)

sahil143 (commented):
Review: data flow redundancies, rendering architecture, CSS co-location, and PatternFly usage.

src/shared/components/pipeline-run-logs/logs/log-viewer-content.ts
```diff
@@ -0,0 +1,26 @@
...
+  sections: readonly LogSection[],
+): PreparedLogViewerContent {
+  if (sections.length === 0) {
+    return { processedData: '', downloadData: '' };
+  }
+
+  const formatted = formatSectionsForSearch(sections);
+  const hasLogContent = sections.some((s) => s.data.length > 0);
+  return {
+    processedData: normalizeLogLines(formatted).join('\n'),
```
sahil143: **suggestion:** `normalizeLogLines` (ANSI stripping + line-ending normalization + split) runs here on the full concatenated log text for search, and then runs again independently in `useSectionRows.ts:36` on each `section.data` for rendering. Both paths process the same raw data.

During WebSocket streaming, sections update frequently — this means the same normalization work runs twice on every update. Consider sharing the normalized result between the search and rendering paths (e.g. normalize once per section, then derive both the search string and `allLines` from the shared result).

src/shared/components/pipeline-run-logs/logs/LogViewer.tsx
```diff
@@ -89,16 +89,19 @@ const LogViewer: React.FC<Props> = ({
   const [logTheme, setLogTheme] = useLogViewerTheme();
   const themeCheckboxId = React.useId();
 
-  // Auto-scroll and resume button logic
+  const { processedData, downloadData } = React.useMemo(
+    () => prepareLogViewerContent(sections),
+    [sections],
+  );
+
+  const lines = React.useMemo(() => processedData.split('\n'), [processedData]);
```
sahil143: **suggestion:** `prepareLogViewerContent` internally calls `normalizeLogLines(formatted).join('\n')` to produce `processedData` as a single string, then this line immediately does `.split('\n')` to get back a `string[]`.

This is a join-then-split roundtrip that allocates two unnecessary intermediate strings on every sections change. Consider having `prepareLogViewerContent` return `string[]` directly instead of joining into a string.

src/shared/components/pipeline-run-logs/logs/log-viewer-content.ts
```diff
@@ -0,0 +1,26 @@
+import { normalizeLogLines } from '~/shared/components/virtualized-log-viewer/log-viewer-utils';
+import type { LogSection } from '~/shared/components/virtualized-log-viewer/types';
+
+export function formatSectionsForSearch(sections: readonly LogSection[]): string {
+  return sections.map((s) => `${s.containerName.toUpperCase()}\n${s.data}`).join('\n\n');
```
sahil143: **suggestion:** `s.containerName.toUpperCase()` is called here in `formatSectionsForSearch`, but the same string is already uppercased at the source in `Logs.tsx:202` (`c.name.toUpperCase()`) and again during rendering in `section-log-ui.tsx:23` (`row.sectionName.toUpperCase()`).

For the `Logs.tsx` path, the same string is uppercased 3 times across the data flow. Consider normalizing case once at section creation and consuming it as-is downstream.

src/shared/components/virtualized-log-viewer/VirtualizedLogViewer.tsx
```diff
@@ -18,49 +18,64 @@ export interface VirtualizedLogViewerProps {
...
 
-  // Prioritize toolbarContext.rowInFocus if it has a valid rowIndex (>= 0)
-  // This allows programmatic navigation via scrollToRow() to work correctly
   const rowInFocus =
     toolbarContext?.rowInFocus && toolbarContext.rowInFocus.rowIndex >= 0
       ? toolbarContext.rowInFocus
       : searchedWordIndexes[currentMatchIndex];
 
-  // Determine effective scroll row based on search or prop
   const effectiveScrollToRow = React.useMemo(() => {
```
sahil143: **suggestion:** `effectiveScrollToRow` re-derives the scroll target from `rowInFocus` (read from toolbar context). But the `scrollToRow` prop already comes from `useLogViewerSearch.scrolledRow`, which performs the same computation: `rowInFocus.rowIndex >= 0 ? rowInFocus.rowIndex + 1 : fallback`.

When search is active, both paths independently arrive at `rowInFocus.rowIndex + 1`. The prop is only useful as fallback for auto-scroll. This creates two sources of truth for scroll position. Consider either:
- Letting `VirtualizedLogViewer` use the prop directly without re-deriving from context, or
- Having it derive exclusively from context and dropping the prop

src/shared/components/virtualized-log-viewer/VirtualizedLogContent.tsx
sahil143: **suggestion:** `<Flex style={{ position: 'relative' }}>` is used purely as a positioning container — no flex layout props are applied. PF v5 Flex adds CSS defaults (`flex-wrap: wrap`, `align-items: baseline`, `row-gap: 0.5rem`) that are not needed here. A plain `<div style={{ position: 'relative' }}>` would be simpler and avoid PF layout overhead.

src/shared/components/virtualized-log-viewer/VirtualizedLogContent.tsx
```diff
@@ -243,55 +233,43 @@ export const VirtualizedLogContent: React.FC<VirtualizedLogContentProps> = ({
...
+            height: `${virtualizer.getTotalSize()}px`,
             width: '100%',
             position: 'relative',
-            display: 'flex',
           }}
         >
-          {/* Line number gutter */}
           <LineNumberGutter
             virtualItems={virtualItems}
             itemSize={itemSize}
```
sahil143: **suggestion — split gutter architecture:** `virtualItems` is iterated twice — once in `LineNumberGutter` (line 247) and once for the content column (line 255). Each visible row produces 2 DOM subtrees.

Rendering `[lineNumber | content]` as a single row element would:
- Halve the DOM nodes per visible row
- Eliminate dual `virtualItems.map()` iteration
- Keep line numbers aligned with wrapped content (currently `LineNumberGutter` uses fixed `height: itemSize` per cell while content uses `measureElement` for dynamic height — when content wraps to multiple visual lines, the gutter cell is shorter than its content counterpart)

The sticky gutter effect can be preserved with `position: sticky; left: 0` on the line number cell within each row, which the SCSS already defines.

src/shared/components/virtualized-log-viewer/section-log-ui.tsx
```diff
@@ -0,0 +1,80 @@
...
+      pointerEvents: 'none',
+      transform: `translateY(${pushUpOffset}px)`,
+      height: `${itemSize}px`,
+      background:
+        'var(--pf-v5-c-log-viewer__main--BackgroundColor, var(--pf-v5-global--BackgroundColor--200))',
+      boxShadow: 'var(--pf-v5-global--BoxShadow--sm)',
+    }}
+    data-test={`sticky-header-${row.sectionName}`}
+  >
+    <Flex
```
sahil143: **suggestion:** A few concerns with this file:

**File naming:** Exports 3 React components but uses kebab-case. Every other component file in this directory uses PascalCase (`VirtualizedLogViewer.tsx`, `SectionedVirtualRow.tsx`, `LineNumberGutter.tsx`).

**No co-located SCSS:** Uses custom classes from other components' SCSS files with no direct import:
- `line-number__gutter`, `line-number__line-number` — defined in `LineNumberGutter.scss`
- `log-content__content-column` — defined in `VirtualizedLogContent.scss`

This is an implicit style dependency — works only because the consumer (`VirtualizedLogContent.tsx`) imports those SCSS files.

**Single-child Flex** (lines 61-68 and 69-78): Both inner `Flex` containers in `StickySectionHeaderBar` wrap a single `<FlexItem>`. Flex layout provides no benefit with one child. For the content column (line 69), the `<FlexItem>` wrapper around `<SectionHeaderButton>` is also unnecessary.

**Unnecessary `position: relative`** (line 65): Nothing inside this Flex uses absolute positioning — `position: relative` has no effect.

src/shared/components/virtualized-log-viewer/SectionedVirtualRow.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const SectionedVirtualRow: React.FC<SectionedVirtualRowProps> = ({
+  virtualIndex,
+  start,
+  row,
+  measureElement,
+  isLineHighlighted,
+  onToggleSection,
+  renderLogLine,
+}) => {
+  const rowClassName = `pf-v5-c-log-viewer__list-item${
```
sahil143, line 48: **nitpick:** `log-content__line--highlighted` is used here but defined in `VirtualizedLogContent.scss`. This component imports no SCSS file, creating an implicit style dependency that would silently break if this component were used outside `VirtualizedLogContent`.

sahil143:
**nitpick** on `SectionedVirtualRow.tsx:33`: `log-content__line--highlighted` is used here but defined in `VirtualizedLogContent.scss`. This component imports no SCSS file, creating an implicit style dependency that would silently break if this component were used outside `VirtualizedLogContent`.

sahil143:
**Directory structure: `virtualized-log-viewer/` is a flat bag of 22 files**

`VirtualizedLogContent.tsx` imports 16 local files. The flat structure gives no signal about which files form a pipeline, which are independent utilities, and which are UI components. As this directory grows with foldable sections, it will get harder to navigate.

The files naturally cluster into 5 orthogonal concerns:

| Concern | Files |
|---|---|
| **Section data** | `useSectionFold`, `useSectionRows` |
| **Section UI** | `section-log-ui.tsx`, `SectionedVirtualRow`, `sticky-section-header` |
| **Syntax highlighting** | `refractor-log`, `useTokenization`, `token-renderer` |
| **Search** | `useSearchRegex`, `useLineRenderer` |
| **Navigation** | `useKeyboardNavigation`, `useLineNumberNavigation`, `useVirtualizedScroll`, `useScrollTop` |

A sub-directory per concern (e.g. `sections/`, `syntax/`, `search/`, `navigation/`, `utils/`) would make imports self-documenting and scope changes to the relevant cluster.

Additionally, `useSectionRows` currently normalizes log lines internally — if normalization were extracted to a shared utility, it could be consumed by both `useSectionRows` and `log-viewer-content.ts` (in `pipeline-run-logs/`), eliminating the duplicate `normalizeLogLines` call noted in the inline comments.

This doesn't need to block this PR. Could you create a follow-up ticket to restructure the directory?

rrosatti (commented):
I found some issues while testing the branch:

1. when you search for something, you're not able to fold/unfold the logs sections:

https://github.com/user-attachments/assets/1b6a7a99-6608-4eb5-853b-02acdc94cd32

2. the line number navigation still doesn't work:

https://github.com/user-attachments/assets/d896d40a-2e6a-4767-af7d-42605c04f11b

could you please take a look on those?

src/components/ImportForm/ComponentSection/ComponentSection.tsx
```diff
@@ -37,6 +38,7 @@ export const ComponentSection = () => {
         label="Component name"
         isRequired
         data-test="component-name"
+        maxLength={MAX_RESOURCE_NAME_LENGTH}
```
sahil143: Why is this change needed in this PR?

rrosatti:
The search issue seems to be fixed! but the line navigation still persists. Maybe we're just missing to unfold the section which the line is part of, because the line itself is being highlighted, we're just missing the unfold -> navigate / scroll to :)

JoaoPedroPP (commented):
Got the same issue as Rodrigo. On Firefox the line number is not working and on safari line number and search are not working.

JoaoPedroPP:
It is working, however I found an inconsistency. When I try to navigate to a specific line I have different outputs when using kubearchive and tekton.

Kubearchive seem to take more time to load and once the line is fulfilled it display its content.
Screenshot 2026-07-03 at 09 52 46

While in tekton I have this
Screenshot 2026-07-03 at 09 54 27

sahil143:
> It is working, however I found an inconsistency. When I try to navigate to a specific line I have different outputs when using kubearchive and tekton.
> 
> Kubearchive seem to take more time to load and once the line is fulfilled it display its content. Screenshot 2026-07-03 at 09 52 46
> 
> While in tekton I have this Screenshot 2026-07-03 at 09 54 27

@JoaoPedroPP That's expected. I am wondering if we should allow line sharing only for kubearchive because tekton results fetched logs are shuffled, it not consistent order of logs between 1st and 2nd api call for the same logs but kubearchive has that because we fetch logs for each container separately.

sahil143:
@coderabbitai Explain data flow in latest changes using diagrams. Explain in detail for when data is fetched from cluster, tekton results or kubearchive.

sahil143 (commented):
Mostly looks good

Image

I am just seeing a blank screen when logs are loading

src/components/ImportForm/ComponentSection/ComponentSection.tsx
janaki29: Not required for logs, just optional improvement. I can move it to any other PR if you want.

src/shared/components/virtualized-log-viewer/SectionLogUI.tsx
```diff
@@ -0,0 +1,74 @@
...
+import './VirtualizedLogContent.scss';
+
+export const SectionHeaderButton: React.FC<{ row: SectionHeaderRow; onToggle: () => void }> = ({
+  row,
+  onToggle,
+}) => (
+  <Button
+    variant="plain"
+    className="pf-v5-u-p-0"
+    style={{ pointerEvents: 'auto' }}
```
rrosatti: nit: we can use css style instead of this inline style :)

src/shared/components/virtualized-log-viewer/useVirtualizedScroll.ts
```diff
@@ -27,10 +32,14 @@ export const useVirtualizedScroll = ({
   virtualizer,
   scrollToRow,
   onScroll,
-}: UseVirtualizedScrollParams) => {
+}: UseVirtualizedScrollParams): UseVirtualizedScrollResult => {
   const prevScrollOffset = React.useRef<number>(0);
   const scrollToIndexRef = React.useRef<number | undefined>(undefined);
 
+  const clearScrollTracking = React.useCallback(() => {
```
rrosatti, line 39: curious question: why do we need this logic? 🤔

rrosatti (commented):
regarding the line navigation logic, I believe we might need to consider waiting for the logs to be fully fetched before navigating. Otherwise, the navigation can fail if it's a high line number (e.g. >100) and the page is still loading the logs 🤔 

also, is it just here or kubearchive is failing quite often while trying to fetch some logs with a `Gateway Time-out` error? 😕

src/shared/components/pipeline-run-logs/logs/log-viewer-content.ts
```diff
@@ -0,0 +1,18 @@
+import type { LogSection, NormalizedLogSection } from '~/shared/components/virtualized-log-viewer/types';
+
+export function formatSectionsForSearch(sections: readonly LogSection[]): string {
```
rrosatti: it seems this function is not being used anywhere, so we can remove it

src/shared/components/virtualized-log-viewer/LineNumberGutter.tsx
```diff
@@ -26,11 +28,13 @@ export const LineNumberGutter: React.FC<LineNumberGutterProps> = ({
   itemSize,
   onLineClick,
   isLineHighlighted,
+  getLineNumber,
 }) => {
   return (
     <Flex className="line-number__gutter" direction={{ default: 'column' }}>
       {virtualItems.map((virtualItem) => {
-        const lineNumber = virtualItem.index + 1;
+        const lineNumber = getLineNumber ? getLineNumber(virtualItem.index) : virtualItem.index + 1;
```
rrosatti: it seems this prop is not being used anywhere, right? so we can undo the change and leave as it was before:

```tsx
const lineNumber = virtualItem.index + 1;
```

src/components/ImportForm/ComponentSection/ComponentSection.tsx
rrosatti: nit: I'd move this change to a separate PR

src/shared/components/pipeline-run-logs/logs/LogViewer.tsx
```diff
@@ -326,9 +342,11 @@ const LogViewer: React.FC<Props> = ({
               <VirtualizedLogViewer
                 key={taskRun?.metadata?.uid || 'default'}
                 sections={sections}
+                normalizedSections={normalizedSections}
```
rrosatti, line 345: do we really need to pass down both `sections` and `normalizedSections` as different props? From what I'm seeing in `VirtualizedLogViewer` -> `VirtualizedLogContent` they are really using only `normalizedSections`, right? So, couldn't we simply pass `normalizedSections` as `sections` prop (`sections={normalizedSections}` - need to fix the prop type) instead of having both props? Let me know if I missed something :)

src/shared/components/virtualized-log-viewer/index.ts
```diff
@@ -2,12 +2,14 @@ export { VirtualizedLogViewer } from './VirtualizedLogViewer';
 export type { VirtualizedLogViewerProps } from './VirtualizedLogViewer';
 export { VirtualizedLogContent } from './VirtualizedLogContent';
 export type { VirtualizedLogContentProps } from './VirtualizedLogContent';
-export type { SearchedWord, LogSection } from './types';
+export type { SearchedWord, LogSection, NormalizedLogSection } from './types';
+export { normalizeSection } from './log-viewer-utils';
 export { LineNumberGutter } from './LineNumberGutter';
```
rrosatti: nit: looks like Fullsend comment makes sense here, it seems we're not using the `LineNumberGutter` component anymore, so we could remove the file and its tests :)

src/shared/components/virtualized-log-viewer/useLineNumberNavigation.ts
```diff
@@ -57,12 +71,26 @@ export const useLineNumberNavigation = (): UseLineNumberNavigationResult => {
...
+  const [highlightedLines, setHighlightedLines] = React.useState<HighlightedLineRange | null>(
+    () => (readyToNavigate ? getHighlightedLines() : null),
+  );
 
   // Track the last hash to detect changes (including when hash is cleared)
   const lastHashRef = React.useRef(window.location.hash);
 
+  // Once navigation becomes ready, apply whatever hash is currently in the URL — it may have
+  // changed while navigation was deferred (e.g. logs were still being fetched).
+  const prevReadyToNavigateRef = React.useRef(readyToNavigate);
```
rrosatti, line 85: nit: do we really need to have this ref logic? can we get it driven by `readyToNavigate` only? 🤔

src/shared/components/virtualized-log-viewer/index.ts
janaki29: updated

src/shared/components/virtualized-log-viewer/useLineNumberNavigation.ts
janaki29: Needed because the hash can be present in the URL before readyToNavigate becomes true, in that case the hash never changes, so only remembering the previous readyToNavigate value (not a hash diff) can detect the "just became ready, must apply now" transition.

src/shared/components/pipeline-run-logs/logs/LogViewer.tsx
janaki29: sections (raw) is still needed because useSectionFold reads section.isCompleted per section to drive fold/auto-expand behaviour, a field NormalizedLogSection doesn't carry, so swapping it in would silently break that.

src/shared/components/pipeline-run-logs/logs/useAutoScrollWithResume.ts
```diff
@@ -1,7 +1,20 @@
 import React from 'react';
 
+interface ActiveLineTarget {
```
rrosatti: let's re-use the `HighlightedLineRange` interface (it's defined in `src/shared/components/virtualized-log-viewer/useLineNumberNavigation.ts`)

rrosatti (commented):
thanks for the changes! I think we're close to get it shipped soon!  :)

but when testing locally, I found two issues:

1. when searching for a text that's part of a folded section, it won't navigate to it. When switching between next/prev, then it navigates to the correct line. Is it the expected behavior? 🤔 

https://github.com/user-attachments/assets/0311ea54-a250-4705-92b2-00ce27f94101

2. live logs are not showing up to date info if you stay focused on task run logs view. If you select another task run and go back to the previous, then it'll show the full data correctly. In the video below you'll notice that on the side panel the task run will be shown as completed (which is true), but the logs are "still loading" - which is not true. Can you check this, please? 

https://github.com/user-attachments/assets/9a406628-8794-4da0-91f8-d72b0c9482a0

src/shared/components/virtualized-log-viewer/VirtualizedLogViewer.tsx
```diff
@@ -1,69 +1,88 @@
...
+  const prevSearchedInputRef = React.useRef(searchedInput);
+  const prevMatchIndexRef = React.useRef(currentMatchIndex);
+
+  React.useEffect(() => {
+    const isNewSearch = prevSearchedInputRef.current !== searchedInput;
+    prevSearchedInputRef.current = searchedInput;
+
+    if (isNewSearch) {
+      prevMatchIndexRef.current = currentMatchIndex;
+      setExpandSearchTargetRow(undefined);
```
rrosatti: the issue related with searching not highlighting the first match + expanding the section, it's because of this line of code. We should set `expandSearchTargetRow` to the first match's row (if it exists). Otherwise, it sets to `undefined`. So, it looks like this:

```tsx
    if (isNewSearch) {
      if (rowInFocus && rowInFocus.rowIndex >= 0) {
        setExpandSearchTargetRow(rowInFocus.rowIndex + 1);
      } else {
        setExpandSearchTargetRow(undefined);
      }
      prevMatchIndexRef.current = currentMatchIndex;
      return;
    }
```

I tested locally and it looks good, but since you've been working more time on it, could you please check if it looks good and doesn't cause any regression on other parts of your PR? :)

rrosatti (commented):
regarding the second issue: it seems it's happening on `main` branch too, so it's not related directly with this PR's changes. I created a separate ticket to take care of it: https://redhat.atlassian.net/browse/KFLUXUI-1506

rrosatti (approved):
nice work!! :) I've just found a small visual bug, but I believe we can handle it after.

(section is hiding a bit of the top of the scrollbar):

https://github.com/user-attachments/assets/4b7aa900-3bb3-4260-ad61-655c8f1cbd8f

sahil143 (approved):

PR #1026, opened 2026-06-18
feat(KFLUXSE-270): conforma refresh indicator

Omeramsc:
## Fixes 
Fixes: https://issues.redhat.com/browse/KFLUXSE-270

depends on #852 

## Description
The Conforma tab uses React Query with staleTime: Infinity for resolved log data,
and WebSocket for real-time TaskRun list updates.
In normal conditions, data stays fresh automatically.
    
However, users need a way to initiate an update manually when:
    
- WebSocket connection drops
- User suspects stale data after extended inactivity
- User wants explicit confirmation that data is current
    
To solve this, this change adds two elements to the ConformaResultsToolbar:
 - "Last checked: X minutes ago" text: shows when the TaskRun list was last fetched/validated
 - Refresh icon button: clicking invalidates the TaskRun list React Query key
    (logs stay cached since they're immutable; only NEW TaskRuns trigger log fetches)
    

## Type of change
- [v] Feature

image
image

## How to test or reproduce?
Run local Konflux and enter the conforma tab on an application.

## Browser conformance: 
- [x] Chrome

## Summary by CodeRabbit

* **New Features**
  * Added live refresh controls to Conforma results, including **“Last checked”**, **“Live updates paused”**, and a manual **Refresh** action, plus a **multi-arch duplicates** toggle.
  * Enhanced Conforma filtering/search to also match by **rule code**, with improved grouped results presentation.

* **Bug Fixes**
  * Improved partial log failure handling: shows an inline warning while still rendering available results (no full fatal error).
  * Improved expanded image/variant display for detail rows (dash when none, tooltips when multiple variants).

* **Style**
  * Updated Conforma results UI styling (muted text and highlighted refresh button).


Omeramsc:
FYI A lot of the comments were reviewes on the old, first, dependant-upon commit. 
Two comments touched on changes within this change, which I just fixed.

src/components/Conforma/ConformaResultsTab/conforma-taskrun-query.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Returns the React Query cache key for the Conforma security TaskRun list.
+ * Mirrors the key that useK8sWatchResource produces internally so that
+ * queryClient.invalidateQueries() targets the correct cache entry.
+ */
+export const buildConformaSecurityTaskRunQueryKey = (
+  namespace: string,
+  applicationName: string,
+): unknown[] =>
```
rrosatti: nit: I'd type it as `QueryKey` (from `@tanstack/react-query`)

src/components/Conforma/ConformaResultsTab/ConformaGroupedTableHeader.ts
```diff
@@ -0,0 +1,14 @@
+export const conformaGroupedTableColumnClasses = {
+  groupKey: 'pf-m-width-40',
+  violations: 'pf-m-width-20',
+  warnings: 'pf-m-width-20',
+  successes: 'pf-m-width-20',
+};
+
+// ponytail: plain array — createTableHeaders adds nothing without sorting
```
rrosatti: I'd remove this comment

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -72,26 +85,53 @@ export const useApplicationConformaResults = (
...
+      namespace?.length
+        ? buildConformaSecurityTaskRunQueryKey(namespace, applicationName)
+        : [],
+    [namespace, applicationName],
   );
 
-  const [securityTaskRuns, taskRunsLoaded, taskRunsError] = useTaskRunsV2(
-    namespace?.length ? namespace : null,
-    { selector },
+  const taskRunClusterQuery = useK8sWatchResource<TaskRunKind[]>(
```
rrosatti: hmm, this won't work well because with `useK8sWatchResource` alone won't fetch data from tekton-results/kubearchive.

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -72,26 +85,53 @@ export const useApplicationConformaResults = (
...
+    watchOptions,
```
rrosatti: `useQuery` provides a `refetch` callback that we could potentially be using as part of the logic in this PR. Have you tried to use it?  https://tanstack.com/query/v4/docs/framework/react/reference/useQuery.

We could potentially update `useTaskRunsV2` to also return the `refetch` cb. I'd suggest to wait for Sahil to review this PR before doing this change, so we can discuss the better approach :) cc: @sahil143

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
Omeramsc: useTaskRunsV2 is still the data source; the second subscription is metadata-only.

Omeramsc:
@rrosatti can you set ok-to-test flag?
@sahil143 will appreciate it if you can get to review it

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
```diff
@@ -101,6 +128,52 @@ export const ConformaResultsToolbar: React.FC<ConformaResultsToolbarProps> = ({
...
+            aria-label="Refresh Conforma results"
+            data-test="conforma-refresh-button"
+            className={
+              refresh.hasLiveUpdatesPaused
+                ? 'conforma-results-toolbar__refresh-button--highlighted'
+                : undefined
+            }
+            icon={<SyncIcon />}
+          >
+            Refresh
```
rrosatti: taking a closer look at the changes, is this "refresh" logic really a requirement for the "Conforma Results Tab" feature? 🤔 I believe that user's won't be on this page waiting for live changes. And in case they wanna refresh the page, they can do so directly from the browser tab. Maybe having the "last checked" info (last time we fetched data in the page - from etcd and/or archives) would be enough?

cc: @Omeramsc @sahil143

Omeramsc: As mentioned in the description, users need a way to initiate an update manually when:
WebSocket connection drops
User suspects stale data after extended inactivity
User wants explicit confirmation that data is current

I think the button is a good approach and QOL feature. it is not a dead-set requirement, but helps mitigating future complaints about stale info.
At this stage, I don't think we should postpone it longer.

src/hooks/useTaskRunsV2.ts
```diff
@@ -56,6 +63,10 @@ export const useTaskRunsV2 = (
     data: clusterResources,
     isLoading: clusterLoading,
     error: clusterError,
+    dataUpdatedAt,
+    isFetching,
+    isWatchDegraded,
+    refetch,
```
sahil143, line 67: refetch is only implemented for cluster resource? dont we need it for archived resources as well? In case a taskrun is removed from etcd and archived to kubearchive/tekton results.

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
```diff
@@ -125,6 +151,52 @@ export const ConformaResultsToolbar: React.FC<ConformaResultsToolbarProps> = ({
...
+          <FlexItem>
+            <Text
+              component="small"
+              className="conforma-results-toolbar__last-checked"
+              data-test="conforma-last-checked"
+            >
+              Last checked: {relativeTime}
+            </Text>
+          </FlexItem>
+        )}
```
sahil143: This should be extracted into a separate component that's wrapped in React.memo. It triggering rerender on every tick where as only the timestamp needs to be updated not the whole toolbar.

src/components/Conforma/ConformaResultsTab/ConformaResultsToolbar.tsx
```diff
@@ -125,6 +151,52 @@ export const ConformaResultsToolbar: React.FC<ConformaResultsToolbarProps> = ({
...
+        <FlexItem>
+          <Button
+            variant={ButtonVariant.link}
+            onClick={refresh.onRefresh}
+            isDisabled={refresh.isRefreshing}
+            aria-label="Refresh Conforma results"
+            data-test="conforma-refresh-button"
+            className={
+              refresh.hasLiveUpdatesPaused
+                ? 'conforma-results-toolbar__refresh-button--highlighted'
```
sahil143: this class should be in ConformaResultsToolbar.scss but it's in ConformaResultsTab.scss

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -117,27 +140,42 @@ export const useApplicationConformaResults = (
...
   const settling = !allSettled;
 
-  const aggregateError = componentsError ?? taskRunsError ?? aggregatedLogError;
+  const fatalError = componentsError ?? taskRunsError;
+
+  React.useEffect(() => {
+    if (aggregatedLogError) {
+      logger.warn('Partial Conforma log fetch failure', { error: aggregatedLogError });
+    }
+  }, [aggregatedLogError]);
```
sahil143, line 169: I don't think this is needed

src/components/Conforma/ConformaResultsTab/conforma-taskrun-query.ts
```diff
@@ -0,0 +1,52 @@
...
+    ],
+  },
+});
+
+/**
+ * Returns the React Query cache key for the Conforma security TaskRun list.
+ * Mirrors the key that useK8sWatchResource produces internally so that
+ * queryClient.invalidateQueries() targets the correct cache entry.
+ */
+export const buildConformaSecurityTaskRunQueryKey = (
```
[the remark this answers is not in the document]
sahil143: This is a valid comment. I dont see this util being used anywhere in this PR.

sahil143 (changes requested):

src/hooks/useTaskRunsV2.ts
Omeramsc: Good catch!. You're right, refetch was only covering the cluster query,  but not the archive...

Fixed: refetch now refreshes both the cluster watch and the KubeArchive query in parallel. Added test coverage for both paths.

src/components/Conforma/ConformaResultsTab/conforma-taskrun-query.ts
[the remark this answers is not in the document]
Omeramsc: right, fixed

src/components/Conforma/ConformaResultsTab/conforma-taskrun-query.ts
```diff
@@ -0,0 +1,33 @@
...
+  watch: true,
+  selector: {
+    matchLabels: {
+      [PipelineRunLabel.APPLICATION]: applicationName,
+      [PipelineRunLabel.PIPELINE_TYPE]: PipelineRunType.TEST,
+    },
+    matchExpressions: [
+      {
+        key: TektonResourceLabel.pipelineTask,
+        operator: 'In' as const,
```
rrosatti: nit: no need for casting `as const` here

src/components/Conforma/ConformaResultsTab/ConformaRefreshIndicator.tsx
```diff
@@ -0,0 +1,79 @@
...
+              className="conforma-results-toolbar__last-checked"
+              data-test="conforma-last-checked"
+            >
+              Last checked: {relativeTime}
+            </Text>
+          </FlexItem>
+        )}
+        {refresh.isRefreshing && (
+          <FlexItem>
+            <Spinner size="sm" aria-label="Refreshing Conforma results" />
```
rrosatti: we can pass down `refresh.isRefreshing` as `isLoading` prop to the `Button` component, so we don't need to handle it separately :)

src/components/Conforma/ConformaResultsTab/ConformaRefreshIndicator.tsx
```diff
@@ -0,0 +1,79 @@
...
+          </FlexItem>
+        )}
+        <FlexItem>
+          <Button
+            variant={ButtonVariant.link}
```
rrosatti: nit: I'm wondering whether it would be better to set it as a `primary` button instead 🤔 

link: 

Image

primary:

Image

hey @pankajshivpuje 👋 do you have/know any kind of design guidelines for such action? At the moment that's how the whole toolbar looks like:

Image

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
```diff
@@ -61,31 +73,50 @@ export const useApplicationConformaResults = (
   const namespace = useNamespace();
   const isKubearchiveLogsEnabled = useIsOnFeatureFlag('kubearchive-logs');
 
+  // Demo-only escape hatch: `?mock=true` renders local fixture data so every
+  // Conforma feature can be showcased without real Conforma TaskRun data.
+  // Persists in sessionStorage so tab navigation does not drop the flag.
+  const isMockMode = isDemoMockMode();
```
rrosatti: hmm, I'm not sure if that's something we want to be bundled in the code. Specially if it's to be used during demo presentations 🤔 I'd remove this. cc: @sahil143

src/hooks/useTaskRunsV2.ts
```diff
@@ -132,6 +143,16 @@ export const useTaskRunsV2 = (
     },
   );
 
+  const kubearchiveRefetch = kubearchiveQuery.refetch;
+
+  const composedRefetch = React.useCallback(async () => {
+    const results = await Promise.all([
+      refetch(),
+      enableKubearchive ? kubearchiveRefetch() : Promise.resolve(),
```
rrosatti: the refetch will only work for live data (etcd) and kubearchive. We also need to handle tekton-results

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.scss
```diff
@@ -7,3 +7,10 @@
     vertical-align: top;
   }
 }
+
+.conforma-grouped-table {
+  &__secondary-text {
```
rrosatti: is this style being used? if so, we need to replace them with the PF v6 equivalent:

`--pf-v5-global--FontSize--sm` -> `--pf-t--global--font--size--sm`
`--pf-v5-global--Color--200` -> `--pf-t--global--text--color--subtle`

src/components/Conforma/ConformaResultsTab/ConformaRefreshIndicator.tsx
Omeramsc: hmm... I feel like putting another big button after all of the text will be a bit weird. at the same time, the current design include multiple different elements. maybe we should change the order? WDYT?

src/components/Conforma/ConformaResultsTab/useApplicationConformaResults.ts
Omeramsc: sorry, missed the leftovers again. fixed.

src/hooks/useTaskRunsV2.ts
Omeramsc: makes sense, fixed and should be working now.

src/components/Conforma/ConformaResultsTab/ConformaResultsTab.scss
Omeramsc: correct, done

src/components/Conforma/ConformaResultsTab/ConformaRefreshIndicator.tsx
sahil143: IMO it should be a plain button. I noticed that the data is fetched every 30 sec for etcd resources, we may want to increase the stale time since there is a button to refresh the data.

sahil143 (commented):
looks good, just one comment around refetch logic for etcd resource

sahil143 (approved):

rrosatti (approved):
LGTM

PR #1046, opened 2026-06-23
feat(KFLUXUI-1378): Adds reusable favicon status badge hook and compositing utility

milantaky:
## Fixes 

https://redhat.atlassian.net/browse/KFLUXUI-1378

## Description

Adds a dynamic favicon status badge for Konflux pages that watch long-running work (pipeline runs, releases). When a status is active, the browser tab favicon shows a small colored dot overlaid on the Konflux icon (GitHub CI-style).

### Architecture

The favicon logic is split into three layers so domain status mapping stays separate from DOM/canvas work:

| Layer | File | Responsibility |
|-------|------|----------------|
| **Utils** | `src/shared/utils/favicon-badge.ts` | Canvas compositing, DOM favicon link management, original-href capture/restore |
| **Color mapping** | `src/shared/utils/status-color-utils.ts` | `STATUS_COLOR` semantic constants and `getStatusColor(status)` — maps `runStatus` → PatternFly hex colors aligned with list-view topology status icons |
| **Hooks** | `src/shared/hooks/` | React lifecycle wiring |

### New API

```ts
// Generic — takes a hex color, no status knowledge
useFaviconStatusBadge(color: string)

// Domain — map status → color, then call the generic hook
useStatusOnFavicon(status: runStatus)
```

`useFaviconStatusBadge` captures the original favicon href on mount and restores it on unmount. `getStatusColor` always returns a concrete hex string, so callers pass a `runStatus` directly — no `null/undefined` handling needed.

**Tests:**
50 unit tests covering:
- Status → color mapping (`STATUS_COLOR`, topology alignment via `runStatusToRunStatus`)
- Canvas compositing and DOM lifecycle (`favicon-badge.ts`)
- Hook acquire/release/cleanup and async cancellation (`useFaviconStatusBadge`)
- Domain hook wiring (`useStatusOnFavicon`)

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

Screenshot 2026-06-23 at 16 38 45
Screenshot 2026-06-23 at 18 56 04
image

## How to test or reproduce?

In`PipelineRunDetailsView.tsx`:

```ts
import { runStatus } from '~/consts/pipelinerun';
import { useStatusOnFavicon } from '~/hooks/useStatusOnFavicon';
import { pipelineRunStatus } from '~/utils/pipeline-utils';

const faviconStatus = React.useMemo(
  () => (pipelineRun ? pipelineRunStatus(pipelineRun) : runStatus.Unknown),
  [pipelineRun],
);

useStatusOnFavicon(faviconStatus);
```

Then:
1. Open a pipeline run detail page (e.g. `/ns/<tenant>/applications/<app>/pipelineruns/<plr-name>`)
2. Confirm the browser tab favicon shows a colored status dot
3. While the run is in progress → blue; on success → green; on failure → red
4. Navigate away → favicon restores to the default Konflux icon

For releases, wire the same hook with status from `getReleaseStatus(release)` / `useReleaseStatus(release)` in `ReleaseDetailsView.tsx`.

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari - Dynamic favicon updates are not supported by WebKit after page load ([WebKit #75877](https://bugs.webkit.org/show_bug.cgi?id=75877)). Manual testing confirmed the badge does not appear in Safari; GitHub’s PR/Actions status favicons behave the same way.
- [ ] Edge

## Summary by CodeRabbit

* **New Features**
  * Added run-status badge rendering on the browser tab favicon, including safe restoration when statuses change or components unmount.
  * Extended document title handling to optionally coordinate favicon badge updates alongside title changes.
  * Introduced a dedicated hook to manage favicon badge lifecycle based on run status.

* **Tests**
  * Added Jest coverage for favicon badge utilities, including status-to-color mapping, canvas composition, async loading, cancellation behavior, and concurrent ownership restoration.
  * Added Jest coverage for the document title and favicon badge hooks’ update/cleanup behavior.


src/hooks/useDocumentTitle.ts
```diff
@@ -1,14 +1,37 @@
...
+};
 
 /**
- * This hook allows you to dynamically update the title of a webpage based on a specified `title` string.
- * @param {string} title Title to update
+ * Dynamically updates the page title and optionally the favicon status badge.
+ * Restores the default title and favicon on unmount.
  */
-export const useDocumentTitle = (title: string): void => {
+export const useDocumentTitle = (title: string, options?: UseDocumentTitleOptions): void => {
```
rrosatti: nit: I'd keep the document title vs favicon concerns separated. The `useDocumentTitle` would stay as it is, and simply handle the document title changes. And the `src/hooks/useFaviconStatusBadge.ts` hook would be responsible for taking care of the favicon changes. 

So, let's say a component wants to make both changes, it would call both hooks. Example:

```tsx
useDocumentTitle('Konflux');
useFaviconStatusBadge(status);
```

let me know what you guys think :)

src/shared/utils/favicon-badge.ts
```diff
@@ -0,0 +1,164 @@
...
+  return link;
+};
+
+const resetToBaselineFavicon = (): void => {
+  getFaviconLink().href = baselineFaviconHref ?? DEFAULT_FAVICON_HREF;
+};
+
+export const getFaviconBadgeColor = (status: string): string => {
+  switch (status) {
+    case 'Succeeded':
```
rrosatti: how about using `runStatus` enum here? (e.g. `runStatus.Succeeded`, etc)

src/shared/utils/favicon-badge.ts
```diff
@@ -0,0 +1,164 @@
...
+      return orangeColor.value;
+    case 'Idle':
+    case 'Pending':
+    case 'Skipped':
+    case 'Unknown':
+    case 'PR needs merge':
+    default:
+      return grayColor.value;
+  }
+};
```
sahil143: This should not be part of favicon status internal implementation.

`useFaviconStatusBadge(color: string)` should take the color instead of status

Then you should have two different hooks
```ts
usePIpelineRunStatusOnFavicon = (status: string) => {
    const color = getColorForPipelineStatus(status);
    useFaviconStatusBadge(color);
}
```
similarly for release

```ts
useReleaseStatusOnFavicon = (status: string) => {
    const color = getColorForReleaseStatus(status);
    useFaviconStatusBadge(color);
}
```

src/shared/utils/favicon-badge.ts
```diff
@@ -0,0 +1,164 @@
...
+    case 'FailedToStart':
+    case 'Test Failures':
+      return redColor.value;
+    case 'Running':
+    case 'In Progress':
+      return blueColor.value;
+    case 'Cancelled':
+    case 'Cancelling':
+      return goldColor.value;
+    case 'Test Warnings':
```
sahil143: don't hardcode strings, use already defined consts/enums for status.

src/hooks/useDocumentTitle.ts
sahil143: +1 👍

sahil143 (changes requested):
move the useFaviconStatusBadge to shared also.

src/shared/utils/favicon-badge.ts
sahil143: 👍

src/shared/utils/favicon-badge.ts
milantaky: updated

src/shared/utils/favicon-badge.ts
milantaky: updated

src/shared/utils/favicon-badge.ts
milantaky: updated

src/hooks/useDocumentTitle.ts
milantaky: updated

rrosatti (commented):
while testing on Firefox, I noticed that sometimes the badge is not visible when switching between different PLR details pages 🤔 

https://github.com/user-attachments/assets/c066ce4a-b9e3-477b-ae4f-967c095c6e25

so far I didn't notice it while testing on Chrome.

it feels like some issue with the acquire/release favicon badge logic 🤔

milantaky:
> while testing on Firefox, I noticed that sometimes the badge is not visible when switching between different PLR details pages 🤔
> 
>  Screen.Recording.2026-06-29.at.11.30.51.mov 
> so far I didn't notice it while testing on Chrome.
> 
> it feels like some issue with the acquire/release favicon badge logic 🤔

Hmmm, will check it out. Nice catch, thank you! @rrosatti

JoaoPedroPP:
Tried on safari and favicon did not change.

Screenshot 2026-06-29 at 08 57 10

src/shared/utils/favicon-badge.ts
```diff
@@ -0,0 +1,134 @@
+const DEFAULT_FAVICON_HREF = '/favicon.ico';
+
+let activeConsumers = 0;
```
rrosatti: I'm wondering: do we really need the `activeConsumers` logic? 🤔 in theory, there's only going to have one consumer at a time, right? (thru `useFaviconStatusBadge`). And we can only be on one detail page at a time.

we could potentially simplify the logic by saving/restoring the baseline directly in the hook. So this utils file would be "stateless"

src/shared/utils/favicon-badge.ts
```diff
@@ -0,0 +1,134 @@
...
+let baselineFaviconHref: string | null = null;
+
+const readFaviconHref = (): string => {
```
rrosatti: we can export this function, so it could be used by `useFaviconStatusBadge`

src/shared/utils/favicon-badge.ts
```diff
@@ -0,0 +1,134 @@
...
+  if (!link) {
+    link = document.createElement('link');
+    link.rel = 'icon';
+    document.head.appendChild(link);
+  }
+  return link;
+};
+
+/** Resets only when no other consumer holds ownership (avoids clobbering active badges). */
+const resetToBaselineFavicon = (): void => {
```
rrosatti: if we go with the "no activeConsumers" approach, then we can delete `resetToBaselineFavicon`, `acquireFaviconBadge`, `releaseFaviconBadge` and `resetFaviconBadgeStateForTests` functions

src/shared/utils/favicon-badge.ts
```diff
@@ -0,0 +1,134 @@
...
+    img.onload = () => resolve(img);
+    img.onerror = () => reject(new Error(`Failed to load favicon image: ${href}`));
+    img.src = href;
+  });
+
+export const applyFaviconBadge = async (
+  color: string,
+  isCancelled?: () => boolean,
+): Promise<void> => {
+  const baseHref = baselineFaviconHref ?? readFaviconHref();
```
rrosatti: nit: `const baseHref = readFaviconHref();`

src/shared/utils/favicon-badge.ts
```diff
@@ -0,0 +1,134 @@
...
+  try {
+    const img = await loadFaviconImage(baseHref);
+    if (isCancelled?.()) {
+      return;
+    }
+
+    const dataUrl = compositeFaviconWithBadge(img, color);
+    if (dataUrl && !isCancelled?.()) {
+      setFaviconHref(dataUrl);
+    } else if (!dataUrl && !isCancelled?.()) {
```
rrosatti: we can get rid off this `else if` if we go with the "no activeConsumers" logic

src/shared/utils/favicon-badge.ts
```diff
@@ -0,0 +1,134 @@
...
+      resetToBaselineFavicon();
+    }
+  } catch {
+    if (!isCancelled?.()) {
```
rrosatti: same here

src/shared/hooks/useFaviconStatusBadge.ts
```diff
@@ -0,0 +1,28 @@
...
+  acquireFaviconBadge,
+  applyFaviconBadge,
+  releaseFaviconBadge,
+} from '~/shared/utils/favicon-badge';
+
+/**
+ * Updates the browser tab favicon with a colored status badge overlay.
+ * Restores the original favicon on unmount.
+ */
+export const useFaviconStatusBadge = (color: string): void => {
```
rrosatti, line 12: as mentioned in a previous message, we could simplify the logic by saving/restoring the baseline directly in this hook:

```tsx
export const useFaviconStatusBadge = (color: string): void => {
  const baselineRef = React.useRef<string | null>(null);

  React.useEffect(() => {
    baselineRef.current = readFaviconHref();
    return () => {
      getFaviconLink().href = baselineRef.current ?? DEFAULT_FAVICON_HREF;
    };
  }, []);

  React.useEffect(() => {
    let cancelled = false;
    void applyFaviconBadge(color, () => cancelled);
    return () => {
      cancelled = true;
    };
  }, [color]);
};
```

rrosatti (dismissed):
I left a few comments for a possible refactor, but honestly the current structure is working just fine!! I tested on Chrome and Firefox and it's working as expected :) 

I'm already approving. If you decided to go with the refactor, just let me know and I'll re-review again after :)

src/utils/status-color-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+    case runStatus.Skipped:
+    case runStatus.Unknown:
+    case runStatus.NeedsMerge:
+    default:
+      return grayColor.value;
+  }
+};
+
+/** Maps release status values (a subset of runStatus) to favicon badge colors. */
+export const getColorForReleaseStatus = (status: runStatus): string => {
```
marcin-michal: couldn't `getColorForReleaseStatus` and `getColorForPipelineStatus` be combined into one? they all map the same statuses into the same colors. it would also allow to combine `usePipelineRunStatusOnFavicon` and `useReleaseRunStatusOnFavicon` into one hook.

sahil143: +1 👍

sahil143 (changes requested):
I would suggest you to define the color for each status instead of directly using the color token.
STATUS_COLOR = {
 success: greencolor.value,
failed: redcolor.value
}
and these color should match the same color which are being used in PipelineRun, TaskRun, Release, Components list pages. I see an overlap between these pages and what you defining from scratch.

src/shared/utils/favicon-badge.ts
milantaky: Actually not, it didn't occur to me that each tab handles it separately 😅 . Thanks for pointing it out!

rrosatti (dismissed):
LGTM. Also tested locally on Chrome and Firefox, both works fine :)

StanislavJochman (dismissed):

sahil143 (approved):

StanislavJochman (approved):
LGTM

PR #1076, opened 2026-06-29
feat(KFLUXUI-1315): add Mintmaker pipeline run list

marcin-michal:
Assisted-by: Cursor

## Fixes 
https://redhat.atlassian.net/browse/KFLUXUI-1312
https://redhat.atlassian.net/browse/KFLUXUI-1314
https://redhat.atlassian.net/browse/KFLUXUI-1315

## Description
Adds dependency update runs from Mintmaker into the component's detail tab, under `Dependency updates` tab.

Also added `scrollMargin` to TanStack Virtual so the virtualizer knows how far the table is offset from the top of the scroll container. Without it, content above the table (headers, tabs, toolbars) causes the virtualizer to miscalculate which rows are visible, producing a visible gap and row cutoff when scrolling. A `ResizeObserver` on the scroll container keeps the margin accurate when the layout above the table changes dynamically.

## Type of change

- [X] Feature
- [ ] Bugfix
- [ ] Build related changes

image

## How to test or reproduce?
Navigate to the Component details page and select the `Dependency updates` tab.

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## Summary by CodeRabbit

* **New Features**
  * Added a feature-gated “Dependency updates” tab and page to component details.
  * Shows dependency pipeline runs in a sortable, filterable table (name/status), with infinite scrolling and distinct empty states for filtered vs unfiltered results.
* **Bug Fixes**
  * Improved consistency in how loading and error states are presented across component and dependency-run data.
  * Refined dependency run row rendering for started time, duration, and status across running, completed, and pending cases.
* **Improvements**
  * Enhanced virtualized table scrolling to account for content above the table for smoother row positioning.


JoaoPedroPP (commented):
LGTM

UI is working as expected. Please take a look in failing checks

sahil143 (changes requested):
@marcin-michal let's use Tablev2 for this change.

marcin-michal:
> @marcin-michal let's use Tablev2 for this change.

updated

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,129 @@
...
+    [sortedDependencyRuns, nameFilter],
+  );
+
+  const { filteredData } = useFilteredData(filterConfigs, nameFilteredRuns, clientFilterValues);
+
+  const statusOptions = React.useMemo(
+    () =>
+      buildOptions(sortedDependencyRuns, (run) => pipelineRunStatus(run), { validKeys: statuses }),
+    [sortedDependencyRuns],
+  );
```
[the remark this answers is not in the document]
marcin-michal: this is correct behaviour

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const [component, componentLoaded, componentError] = useComponent(namespace, componentName, true);
+  const [
+    dependencyRuns,
+    dependencyRunsLoaded,
+    dependencyRunsError,
+    getNextPage,
+    { isFetchingNextPage, hasNextPage },
+  ] = usePipelineRunsV2(
+    componentLoaded && !componentError ? MINTMAKER_NAMESPACE : null,
```
rrosatti, line 41: dummy question: do you know if the MintMaker namespace name will always be `mintmaker` throughout all konflux clusters, or could there be some cases where the it'd have a custom/different name? 👀

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+  );
+
+  const error = componentError ?? dependencyRunsError;
+  if (error) {
+    const loaded = componentError ? componentLoaded : dependencyRunsLoaded;
+    return getErrorState(error, loaded, 'dependency runs');
+  }
+
+  return (
+    <Flex direction={{ default: 'column' }}>
```
rrosatti, line 77: I'm seeing some "weird" layout behavior here. Initially I thought it could be because of `Flex` component, but even after removing it, it behaves the same 🤔 :

https://github.com/user-attachments/assets/5135437b-4f7e-480c-9b11-cb6279e540e2

could you please take a look?

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+          fetchNextPage={() => {
+            if (hasNextPage && !isFetchingNextPage) {
+              getNextPage?.();
+            }
+          }}
+          columnStateKey={DEPENDENCY_RUNS_COLUMN_STATE_KEY}
+        />
+      </TableContainer>
+      {isFetchingNextPage ? (
+        <Stack style={{ marginTop: 'var(--pf-v5-global--spacer--md)' }} hasGutter>
```
rrosatti: `--pf-v5-` class names don't exist anymore after PF v6 migration, so use `--pf-t--global--spacer--md` instead :)

src/feature-flags/flags.ts
```diff
@@ -142,6 +142,12 @@ const InternalFLAGS = {
       visibleInFeatureFlagPanel: false,
     },
   },
+  mintmaker: {
+    key: 'mintmaker',
+    description: 'Enable Mintmaker in the UI',
```
rrosatti: nit: maybe it's MintMaker instead?

rrosatti: also, how about updating the description to be something like: `"Show MintMaker dependency updates tab"`. I think it'd give a clearer explanation what the feature flag will toggle on/off :)

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
marcin-michal: yes, it is `mintmaker` for all clusters

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
marcin-michal: aah, missed it, thank you

src/feature-flags/flags.ts
marcin-michal: as part of this epic, there will be one more cluster-wide page for Mintmaker schedules, so I think a broader name is more fitting

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
marcin-michal: oh, this didn't happen on Firefox :sweat_smile:

updated the new table's `useVirtualization` to take `scrollMargin` prop, which should fix the miscalculation of where the user is scrolled

src/components/Components/ComponentDependencyManager/dependency-runs-table-config.tsx
```diff
@@ -0,0 +1,83 @@
...
+    param: 'status',
+    label: 'Status',
+    filterFn: (item, values) => values.includes(pipelineRunStatus(item)),
+  },
+] as const);
+
+export const dependencyRunsColumns: ColumnDefinition<PipelineRunKind>[] = [
+  {
+    id: 'name',
+    header: 'Name',
```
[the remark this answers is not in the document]
marcin-michal: this is intentended, there are no details we would want to show about the mintmaker plr

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
rrosatti: the issue is fixed now!

rrosatti, line 77: do we really need to wrap it with the `Flex` component? can't we use a fragment (`<>..</>`) instead? 🤔

src/shared/components/TableV2/Table.tsx
```diff
@@ -89,6 +89,28 @@ export const Table = <TData,>({
...
+      return;
+    }
+
+    const recalculate = () => {
+      const tableRect = tableNode.getBoundingClientRect();
+      const scrollRect = scrollElement.getBoundingClientRect();
+      setScrollMargin(tableRect.top - scrollRect.top + scrollElement.scrollTop);
+    };
+
+    recalculate();
```
rrosatti: hmm, I believe we don't need to call this `recalculate()` function right away 🤔

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
marcin-michal: that was just me trying to please the auto review bots, as they sometimes complain, that Patternfly components should be used instead of raw html, as per repo guidelines :D

src/shared/components/TableV2/Table.tsx
marcin-michal: you're right, updated

rrosatti (dismissed):
LGTM!

janaki29 (dismissed):
lgtm

src/shared/components/TableV2/Table.tsx
```diff
@@ -89,6 +89,26 @@ export const Table = <TData,>({
...
+      const scrollRect = scrollElement.getBoundingClientRect();
+      setScrollMargin(tableRect.top - scrollRect.top + scrollElement.scrollTop);
+    };
+
+    const observer = new ResizeObserver(recalculate);
+    observer.observe(scrollElement);
+    observer.observe(tableNode);
+
+    return () => observer.disconnect();
+  }, [tableNode, scrollElement]);
```
sahil143: What is this useEffect fixing?

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { Table, TableContainer } from '~/shared/components/TableV2';
+import { useNamespace } from '~/shared/providers/Namespace';
+import { getErrorState } from '~/shared/utils/error-utils';
+import { PipelineRunKind } from '~/types';
+import { statuses } from '~/utils/commits-utils';
+import { pipelineRunStatus } from '~/utils/pipeline-utils';
+import {
+  DEPENDENCY_RUNS_COLUMN_STATE_KEY,
+  dependencyRunsColumns,
+  filterConfigs,
```
sahil143: Rename this to `dependencyRunsFilterConfig`

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+} from '~/shared/components/Filter';
+import { Table, TableContainer } from '~/shared/components/TableV2';
+import { useNamespace } from '~/shared/providers/Namespace';
+import { getErrorState } from '~/shared/utils/error-utils';
+import { PipelineRunKind } from '~/types';
+import { statuses } from '~/utils/commits-utils';
+import { pipelineRunStatus } from '~/utils/pipeline-utils';
+import {
+  DEPENDENCY_RUNS_COLUMN_STATE_KEY,
+  dependencyRunsColumns,
```
sahil143:
```suggestion
  dependencyRunsTableColumns,
```

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+    dependencyRunsError,
+    getNextPage,
+    { isFetchingNextPage, hasNextPage },
+  ] = usePipelineRunsV2(
+    componentLoaded && !componentError ? MINTMAKER_NAMESPACE : null,
+    React.useMemo(
+      () => ({
+        selector: {
+          filterByCreationTimestampAfter: component?.metadata?.creationTimestamp,
+          filterByName: nameFilter || undefined,
```
sahil143, line 46:
```
          filterByName: nameFilter || undefined,
```
above you have nameFilter = filterValue.name ?? '', here you have it set to undefined in an OR operation. It is redundant.

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const sortedDependencyRuns = React.useMemo((): PipelineRunKind[] => {
+    if (!dependencyRuns) {
+      return [];
+    }
+
+    return [...dependencyRuns].sort((a, b) =>
+      String(b.status?.startTime || '').localeCompare(String(a.status?.startTime || '')),
+    );
+  }, [dependencyRuns]);
```
sahil143: Why is it sorted here? This should be part of sorting config. default sort by a specific column on column config. you can remove this and we can fix this in a separate PR to add default sort column functionality.

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const { filteredData } = useFilteredData(filterConfigs, sortedDependencyRuns, clientFilterValues);
+
+  const statusOptions = React.useMemo(
+    () =>
+      buildOptions(sortedDependencyRuns, (run) => pipelineRunStatus(run), { validKeys: statuses }),
+    [sortedDependencyRuns],
+  );
```
sahil143: I think we should switch the pipeline status filter to use static keys going forward, since the full list of possible statuses is already known.

Right now, the dropdown is inconsistent because it only shows statuses present in the current batch of loaded data. Moving to static keys would fix this by allowing us to properly fetch data on demand (e.g., continuing to paginate until a selected status is found or an empty state is reached). Plus, as we load further runs, we'll naturally accumulate more data for the other statuses anyway. 
It will also ensure the filtering behavior remains identical and predictable across all pipeline run pages.

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+        data={filteredData}
+        unfilteredData={sortedDependencyRuns}
+        loaded={isFetchingNextPage || dependencyRunsLoaded}
+        emptyState={<FilteredEmptyState onClearFilters={clearAll} />}
+        noDataState={<PipelineRunEmptyStateV2 />}
+        toolbar={
+          isFiltered || sortedDependencyRuns.length > 0 ? (
+            <FilterToolbar configs={filterConfigs} options={{ status: statusOptions }} />
+          ) : undefined
+        }
```
sahil143, line 91: Why is the toolbar dependent on data and `isFiltered`. it shouldn't be.

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+    const loaded = componentError ? componentLoaded : dependencyRunsLoaded;
+    return getErrorState(error, loaded, 'dependency runs');
+  }
+
+  return (
+    <Flex direction={{ default: 'column' }}>
+      <TableContainer
+        data={filteredData}
+        unfilteredData={sortedDependencyRuns}
+        loaded={isFetchingNextPage || dependencyRunsLoaded}
```
sahil143: Here isFetchingNextPage shouldn't be used because it will show full table loading indicator when nextpage is being fetched

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+          getRowId={(row) => row.metadata?.uid ?? row.metadata?.name ?? ''}
+          aria-label="Dependency run list"
+          enableSorting
+          hasNextPage={hasNextPage}
+          isFetchingNextPage={isFetchingNextPage}
+          fetchNextPage={() => {
+            if (hasNextPage && !isFetchingNextPage) {
+              getNextPage?.();
+            }
+          }}
```
sahil143: should use `useCallback` or just pass

```suggestion
          fetchNextPage={getNextPage}
```

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,124 @@
...
+      </TableContainer>
+      {isFetchingNextPage ? (
+        <Stack style={{ marginTop: 'var(--pf-t--global--spacer--md)' }} hasGutter>
+          <Bullseye>
+            <Spinner size="lg" aria-label="Loading more pipeline runs" />
+          </Bullseye>
+        </Stack>
+      ) : null}
+    </Flex>
+  );
```
sahil143: Not needed. It's being handled internally in the table component.

src/feature-flags/flags.ts
```diff
@@ -142,6 +142,12 @@ const InternalFLAGS = {
       visibleInFeatureFlagPanel: false,
     },
   },
+  mintmaker: {
+    key: 'mintmaker',
+    description: 'Enable MintMaker in the UI',
+    defaultEnabled: false,
+    status: 'wip',
+  },
```
sahil143, line 150: Do we need a condition for this? How will this work in upstream? By default, would we be able to fetch mintmaker pipeline on any upstream clusters?

src/routes/page-routes/components.tsx
```diff
@@ -29,6 +31,14 @@ const componentRoutes = [
         path: `activity`,
         element: <ComponentActivityTab />,
       },
+      {
+        path: `dependency-updates`,
```
sahil143:
```suggestion
        path: `dep-updates`,
```

src/routes/page-routes/components.tsx
```diff
@@ -29,6 +31,14 @@ const componentRoutes = [
...
+        loader: () => {
+          ensureFeatureFlagOnLoader('mintmaker');
+          return null;
+        },
+        element: <ComponentDependencyTab />,
```
sahil143: Load this dynamically?

sahil143 (changes requested):

src/feature-flags/flags.ts
marcin-michal: yes, it should be available in all clusters

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
marcin-michal: when the list is not filtered and it is empty, there is no point in showing the filter toolbar, as changing it will have no effect

src/shared/components/TableV2/Table.tsx
marcin-michal: it's fixing a bug Rodrigo reported, where there would be an empty gap at the top of the table when scrolled down https://github.com/konflux-ci/konflux-ui/pull/1076#discussion_r3587613893

I added a comment explaining it

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,111 @@
...
+      [component?.metadata?.creationTimestamp, componentName, nameFilter, namespace],
+    ),
+  );
+
+  const dependencyRunsList = React.useMemo(
+    (): PipelineRunKind[] => dependencyRuns ?? [],
+    [dependencyRuns],
+  );
+
+  const { filteredData } = useFilteredData(
```
rrosatti, line 59: somehow the filtering is a bit off, I'm searching for a text that doesn't exist and it's still returning 1 result 👀 

https://github.com/user-attachments/assets/6c724bb7-1b8f-49af-a2ea-bddcb8727a5b

marcin-michal: hmm, good catch, but it seems like this is a bigger issue, and should be probably fixed in a different PR :sweat_smile: 

The reason why this one PLR is showing is because it's on the cluster and not archived yet. Seems like Kubernetes doesn't support comparing by substrings, so the `filterByName` is basically lost in the query to the live cluster. Why this wasn't an issue on for example `PipelineRunsListView.tsx` is because it was filtered by name again on the UI side https://github.com/konflux-ci/konflux-ui/blob/main/src/components/PipelineRun/PipelineRunListView/PipelineRunsListView.tsx#L141-L144

This list uses the TableV2 with its filter config and since `name` is set to `api` mode, there's no additional filtering done on the UI side, that would catch this.

Since this will be an issue in any list filtering by name with the new table, I think its better to fix it separately.

cc: @sahil143

rrosatti: got it! thanks for the explanation! so I believe it might be better to create a separate ticket for this issue :)

rrosatti (dismissed):
code looks good to me. Since the bug/issue found is not directly related to this PR's changes, I think we can merge it and fix the issue separately :) great work!

janaki29 (dismissed):

src/shared/components/TableV2/TableBody.tsx
```diff
@@ -48,17 +50,18 @@ export const TableBody = <TData,>({
   expandedContent,
   visibleColumnCount,
   isFetchingNextPage,
+  scrollMargin = 0,
 }: TableBodyProps<TData>) => {
   const lastVirtualRow = virtualRows[virtualRows.length - 1];
   const bottomSpacerHeight = lastVirtualRow
```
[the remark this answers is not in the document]
marcin-michal: makes no difference

StanislavJochman (dismissed):
LGTM

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,111 @@
...
+        },
+      }),
+      [component?.metadata?.creationTimestamp, componentName, nameFilter, namespace],
+    ),
+  );
+
+  const dependencyRunsList = React.useMemo(
+    (): PipelineRunKind[] => dependencyRuns ?? [],
+    [dependencyRuns],
+  );
```
sahil143: This is not needed

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const { filteredData } = useFilteredData(
+    dependencyRunsFilterConfig,
+    dependencyRunsList,
+    clientFilterValues,
+  );
+
+  const statusOptions: FilterOption[] = React.useMemo(
+    () => statuses.map((s) => ({ label: s.charAt(0).toUpperCase() + s.slice(1), value: s })),
```
sahil143: use capitalize util for this

src/components/Components/ComponentDependencyManager/DependencyRunsListView.tsx
sahil143: but if user has some filter enabled and nothing matched that filter, even in that case it would hide the filter bar?

sahil143 (changes requested):

marcin-michal:
> but if user has some filter enabled and nothing matched that filter, even in that case it would hide the filter bar?

@sahil143 the filter is present if either the result is filtered or the non-filtered list is not empty, so no

src/shared/components/TableV2/Table.tsx
sahil143: This is a good improvement. encapsulate it into a separate hook useScrollMargin that returns the calculated scrollMargin.

sahil143 (changes requested):

rrosatti (dismissed):
LGTM

sahil143 (approved):

rrosatti (approved):

StanislavJochman (approved):
LGTM

PR #1077, opened 2026-06-29
fix(KFLUXSE-166): fix Conforma Results Tab compatibility with KubeArc…

Omeramsc:
## Fixes 
https://redhat.atlassian.net/browse/KFLUXSE-312

## Description
The Conforma Results Tab showed 0 results when kubearchive-logs was enabled, and rendered a 404-like error page when taskruns-kubearchive and pipelineruns-kubearchive were additionally enabled.

Two root causes:

1. fetchConformaLogFromKubearchive used commonFetchJSON, but the KubeArchive pod log endpoint returns text/plain, causing JSON parse failures that silently zeroed out all results.

2. resolveConformaResultFromTaskRun propagated KubeArchive errors (including the common case of archived TaskRuns lacking status.podName) directly to React Query, which eventually surfaced as a 404-like error state when all component queries failed.

Fix:
- Add a three-tier fallback to fetchConformaLogFromKubearchive: commonFetchJSON -> commonFetchText + JSON.parse -> extractConformaResultsFromTaskRunLogs.
- Change resolveConformaResultFromTaskRun to return a tagged ConformaFetchResult instead of a bare ConformaResult. On KubeArchive failure (fetch error or missing podName), it falls back to Tekton Results and marks kubearchiveFailed: true.
- Expose kubearchiveFailedCount from useApplicationConformaResults and show an inline warning Alert in ConformaResultsTab so users know data is sourced from Tekton Results when KubeArchive is unavailable.

Assisted-by: Claude

## Type of change
- [ ] Feature
- [v] Bugfix
- [ ] Build related changes

## Summary by CodeRabbit

* **New Features**
  * Added a data source notice in Conforma results when some logs can’t be fetched from KubeArchive.
  * Results now clearly show when they are coming from an alternate source.

* **Bug Fixes**
  * Improved log retrieval so results can fall back to another source if the primary fetch fails.
  * Better handling for mixed or unexpected log response formats, helping results still load when possible.


src/components/Conforma/ConformaResultsTab/ConformaResultsTab.tsx
```diff
@@ -117,6 +119,17 @@ const ConformaResultsTabContent: React.FC = () => {
...
   return (
     <>
+      {kubearchiveFailedCount > 0 && (
+        <Alert
+          variant="warning"
+          isInline
+          title="Data source notice"
+          data-test="kubearchive-fallback-warning"
+        >
+          KubeArchive log fetch failed for {kubearchiveFailedCount} component
```
rrosatti, line 129: it sounds like a nice approach: if kubearchive fails, fetch from tekton-results. But, I think if we want to have such feature, it should be also applied to other resources (not only logs) and it'd be a bigger work :) (probably involving UX as well, to find a good way to communicate such thing)

I don't know if such approach was already discussed in the past and/or if it's something that we do (not) want to do. @sahil143 do you know?

PR #1136, opened 2026-07-09
fix(KFLUXUI-1284): improve opaque build secret create-or-use flow

milantaky:
## Fixes 

https://redhat.atlassian.net/browse/KFLUXUI-1284

## Description

**UX approved**
Improves the build secret modal so users can `create` a new opaque secret or `reuse` an existing cluster secret from the same flow.

Updates Build-time Secret modal header to `Create or use new build secret`

## Type of change

- [x] Feature
- [x] Bugfix
- [ ] Build related changes

https://github.com/user-attachments/assets/b23475d9-a367-4b6b-a154-2d6fea54bb10

## How to test or reproduce?

1. `Applications` -> select application -> `Add component` 
2.  Fill in git repository (no need to be valid)
3. In section `Build time secret` click `Add secret`

- Key/value secret shows list of existing secrets
  - Create new -> `no change`
  - Select existing -> `not editable` (except name) -> confirm button says `Use`
  - Select existing  -> edit name:
    - Name is unique -> Secret values `kept`, but now `editable`, button says `Create`
    - Name is matching other existing secret -> Values change to new secret
- Image pull secret - `no change`
- Source secret - `no change`

Verify that the secrets are created when component is created.

## Browser conformance: 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## Summary by CodeRabbit

* **New Features**
  * Added an edit mode for secrets that locks secret type/name and adjusts form behavior.
  * Improved opaque-secret selection with a searchable dropdown; selecting an existing opaque secret now pre-fills key/value and labels and adapts the modal title and submit actions.

* **Bug Fixes**
  * Refined opaque-secret synchronization for read-only keys/labels, correct repopulation when switching selections/types, and better handling of label/value defaults.
  * Enhanced validation and form controls so “use existing” correctly permits/blocks empty opaque key/value pairs; capped dropdown menu height; disabled add key/value when reusing.


src/components/Secrets/hooks/opaqueSecretSyncUtils.ts
```diff
@@ -0,0 +1,101 @@
...
+      keyValues: [...getSupportedPartnerTaskKeyValuePairs(secretName)],
+      labels: DEFAULT_OPAQUE_LABELS,
+    };
+  }
+
+  const matched = findExistingOpaqueSecretByName(secretName, existingSecrets);
+  return {
+    keyValues: matched?.opaque?.keyValuePairs,
+    labels: matched?.labels?.length ? matched.labels : DEFAULT_OPAQUE_LABELS,
+  };
```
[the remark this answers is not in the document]
JoaoPedroPP: Does this make sense?

_Edit: I think you have the checker below._

src/components/Secrets/hooks/useOpaqueSecretSync.ts
```diff
@@ -0,0 +1,114 @@
...
+      case 'makeEditable':
+        makeOpaqueFieldsEditable();
+        break;
+      case 'resetOpaque':
+        resetOpaqueFields();
+        break;
+      case 'none':
+        break;
+      default:
+        break;
```
JoaoPedroPP, line 91: nit: `case: 'none'` is `default`

JoaoPedroPP (commented):
Overall LGTM

Just check if one of the fullsend comment make sense and a small nit up to you to check.
Locally the UI is working as expected.

src/components/ImportForm/SecretSection/SecretSection.tsx
```diff
@@ -154,14 +168,14 @@ const SecretSection: React.FC<SecretSectionProps> = ({ currentComponent }) => {
                 <Flex spaceItems={{ default: 'spaceItemsNone' }}>
```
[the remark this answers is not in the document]
milantaky: Leaving as-is. isAriaDisabled is intentional so the tooltip still works (same pattern as ButtonWithAccessTooltip). PF replaces onClick when aria-disabled, and we already assert the modal does not open for cluster secrets.

src/components/Secrets/hooks/useOpaqueSecretSync.ts
milantaky: I only merged the case none and default for readability. When removing the default, linter has problems and I dont want to post supress comments. Thanks for nitting!

src/components/Secrets/__tests___/SecretForm.spec.tsx
JoaoPedroPP, line 1: I am not against to use `getByTestId` when necessary but we should get and query, as most as we can, through elements that are in the screen instead of id tags. Please evaluate if possible to reduce usage of `getByTestId`.

src/shared/components/formik-fields/key-value-file-input-field/__tests__/KeyValueFileInputField.spec.tsx
```diff
@@ -115,3 +123,36 @@ test('should add new entry on clicking Add key/value button', async () => {
...
+      onSubmit={onSubmit}
+      disableAddAction
+      initialValues={{
+        keyValueInput: [{ key: 'key-three', value: 'value-three' }],
+      }}
+    />,
+  );
+
+  await waitFor(() => {
+    expect(screen.queryByTestId('add-key-value-button')).not.toBeInTheDocument();
```
JoaoPedroPP, line 139: Can we query a visual element instead an id?

JoaoPedroPP (commented):
LGTM

UI is working as expected. Left a comment about test approach.

src/shared/components/formik-fields/key-value-file-input-field/__tests__/KeyValueFileInputField.spec.tsx
milantaky: updated:)

JoaoPedroPP: I am still seeing `queryByTestId` in the file. is it possible to use an alternative?

src/components/Secrets/hooks/opaqueSecretSyncUtils.ts
```diff
@@ -0,0 +1,109 @@
...
+      };
+    }
+
+    return {
+      keyValues: [...getSupportedPartnerTaskKeyValuePairs(secretName)],
+      labels: DEFAULT_OPAQUE_LABELS,
+    };
+  }
+
+  const matched = findExistingOpaqueSecretByName(secretName, existingSecrets);
```
rrosatti, line 48: nit: I see that we're duplicating the same logic here and inside the `isUsingExistingClusterSecret(secretName, SecretTypeDropdownLabel.opaque, existingSecrets)` if statement. How about simplify to something like:

```tsx
export const getOpaqueFieldsFromExistingSecret = (
  secretName: string,
  existingSecrets: BuildTimeSecret[],
): { keyValues?: KeyValueEntry[]; labels: KeyValueEntry[] } => {
  if (
    isPartnerTask(secretName, supportedPartnerTasksSecrets) &&
    !isUsingExistingClusterSecret(secretName, SecretTypeDropdownLabel.opaque, existingSecrets)
  ) {
    return {
      keyValues: [...getSupportedPartnerTaskKeyValuePairs(secretName)],
      labels: DEFAULT_OPAQUE_LABELS,
    };
  }

  const matched = findExistingOpaqueSecretByName(secretName, existingSecrets);
  return {
    keyValues: matched?.opaque?.keyValuePairs,
    labels: matched?.labels?.length ? matched.labels : DEFAULT_OPAQUE_LABELS,
  };
};
```

no big deal, just a suggestion. code works as is :)

rrosatti (commented):
curious question: is this supposed to work only during the "create component -> build time secret" flow? Or should it also work when creating a new secret thru Secrets list page? 🤔

src/components/Secrets/hooks/opaqueSecretSyncUtils.ts
milantaky: Heyy, good catch, updated, thank you!
Regarding the question. This is only for the "create component -> build time secret" flow, since when creating a new secret, only `snyk-secret` is displayed, I would keep that as is :)

rrosatti: thank you! yeah, makes sense! :) 

could you please take a look at the `[logic-error]` feedback from Fullsend and check whether it's something that should be fixed or not? 🙏

src/shared/components/formik-fields/SelectInputField.tsx
```diff
@@ -245,6 +245,8 @@ const SelectInputField: React.FC<React.PropsWithChildren<SelectInputFieldProps>>
         onSelect={onSelect}
         onOpenChange={setIsOpen}
         toggle={toggle}
+        isScrollable
+        maxMenuHeight="40vh"
```
rrosatti, line 249: nit: is this really needed? 👀

rrosatti (approved):

StanislavJochman (approved):
LGTM: Just UI question: Would not it make sense to either hide the fields on existing secret / show there something informative - Like if that column is used show something like hidden and keep rest blank? cc @milantaky

src/shared/components/formik-fields/SelectInputField.tsx
milantaky: Without this, the list of secrets (if overflowing the page) could not be scrolled. For example if you wanted to select the bottom secret :)

PR #1167, opened 2026-07-14
fix(KFLUXUI-1507): handle grant access step in stage login flow

rrosatti:
## Fixes 

Fixes: https://issues.redhat.com/browse/KFLUXUI-1507

## Description

Handle OAuth "Authorize Access" and Dex "Grant Access" consent pages that now appear during stage login. Both steps are conditional since they may be skipped depending on session state. Also makes the SSO redirect workaround non-blocking.

## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Build related changes

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

## Summary by CodeRabbit

* **Tests**
  * Updated the end-to-end login page object with additional selectors for the “Grant Access” action.
  * Improved the login flow to conditionally handle an optional OAuth consent/approval step when present.
  * Extended the stage login flow to perform “Grant Access” before continuing with the existing workaround and navigation behavior.

### DIFF
--- a/e2e-tests/utils/Login.ts
+++ b/e2e-tests/utils/Login.ts
@@ -76,14 +76,14 @@ export class Login {
     cy.get(stageLoginPO.username).type(username);
     cy.get(stageLoginPO.password).type(password, { log: false });
     cy.get(stageLoginPO.loginButton).click();
+
     // Click through OAuth consent page if it appears
-    cy.get(`${stageLoginPO.approveButton}, ${stageLoginPO.grantAccessClass}`, { timeout: 30000 })
-      .first()
-      .then(($el) => {
-        if ($el.is(stageLoginPO.approveButton)) {
-          cy.wrap($el).click();
-        }
-      });
+    cy.get('body', { timeout: 30000 }).then(($body) => {
+      if ($body.find(stageLoginPO.approveButton).length > 0) {
+        cy.get(stageLoginPO.approveButton).click();
+      }
+    });
+    // Grant Access is always required
     cy.contains(stageLoginPO.grantAccessClass, stageLoginPO.grantAccessText).click();
 
     // ----- Workaround -----
--- a/e2e-tests/utils/Login.ts
+++ b/e2e-tests/utils/Login.ts
@@ -76,15 +76,25 @@ export class Login {
     cy.get(stageLoginPO.username).type(username);
     cy.get(stageLoginPO.password).type(password, { log: false });
     cy.get(stageLoginPO.loginButton).click();
+    // Click through OAuth consent page if it appears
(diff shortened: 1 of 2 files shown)


e2e-tests/utils/Login.ts
```diff
@@ -72,15 +76,25 @@ export class Login {
     cy.get(stageLoginPO.username).type(username);
     cy.get(stageLoginPO.password).type(password, { log: false });
     cy.get(stageLoginPO.loginButton).click();
+    // Click through OAuth consent page if it appears
+    cy.get(`${stageLoginPO.approveButton}, ${stageLoginPO.grantAccessClass}`, { timeout: 30000 })
```
Katka92: I found this step a little confusing - waiting here for the gratAccess and then searching for it again on line 87. We usually wait for the body and check if the element is present, something like this:
```
    // Optional intermediate approve step
    cy.get('body', { timeout: 30000 }).then(($body) => {
      if ($body.find(stageLoginPO.approveButton).length > 0) {
        cy.get(stageLoginPO.approveButton).click();
      }
    });
    // Grant Access is always required
    cy.contains(stageLoginPO.grantAccessClass, stageLoginPO.grantAccessText).click();
```
Although it seems to me like a cleaner solution, I understand it's a matter of preference, so I leave the final decision to you.

e2e-tests/utils/Login.ts
```diff
@@ -72,15 +76,25 @@ export class Login {
...
     // Sometimes page doesn't go to homepage but shows
     // "Something went wrong - Invalid token" message.
-    cy.contains('You have successfully logged into Red Hat Internal SSO', {
-      timeout: 10000,
-    }).should('be.visible');
-    cy.wait(2000);
-    cy.visit(Cypress.env('KONFLUX_BASE_URL'));
+    cy.get('body', { timeout: 10000 }).then(($body) => {
+      if ($body.text().includes('You have successfully logged into Red Hat Internal SSO')) {
+        cy.wait(2000);
```
Katka92, line 94: Why do we wait here?

rrosatti: to be completely hones, I don't know 😄 I just kept the previous logic, which was already calling `cy.wait(2000);` :)

e2e-tests/utils/Login.ts
rrosatti: I like the suggestion, I'll give it a shot!

StanislavJochman (approved):

e2e-tests/utils/Login.ts
Katka92: Hm, ok, let's keep it there. I again missed it's just a styling change... 🤦‍♀️

Katka92 (approved):

PR #1175, opened 2026-07-16
fix(KONFLUX-14741): handle kubearchive pod/container 404 in log viewer

janaki29:
## Fixes 
[KONFLUX-14741](https://redhat.atlassian.net/browse/KONFLUX-14741)

## Description

- When some container/step log fetches fail and others succeed, keep showing available logs and surface the failure on that step (message in the expanded section body).

- For archived pods (kubearchive-logs), always fetch via HTTP/KubeArchive and never open a live websocket against a pod that no longer exists.

- When KubeArchive is enabled, show a clear error if the pod cannot be loaded.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

## Browser conformance: 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

[KONFLUX-14741]: https://redhat.atlassian.net/browse/KONFLUX-14741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Log errors are now shown per container or step instead of as one global error.
  * Failed log sections display a warning indicator, tooltip, and inline failure message.
  * Failed sections remain expanded for easier troubleshooting, while successful sections continue normal folding behavior.
  * Archived log sources now reliably use archived log retrieval without attempting live streaming.

* **Bug Fixes**
  * A failed container no longer hides logs successfully retrieved from other containers.
  * Missing archived logs are treated as empty results without displaying misleading fetch errors.


src/shared/components/virtualized-log-viewer/SectionLogUI.tsx
```diff
@@ -24,6 +25,19 @@ export const SectionHeaderButton: React.FC<{ row: SectionHeaderRow; onToggle: ()
...
       </FlexItem>
+      {row.error && (
+        <FlexItem>
+          <Tooltip content={row.error}>
+            <Icon
+              status="danger"
+              data-test={`fold-header-error-${row.sectionName}`}
+              aria-label={`Failed to fetch logs for this step: ${row.error}`}
+            >
+              <ExclamationCircleIcon aria-hidden />
```
marcin-michal: this should have some small padding on the right and be centered vertically in the header

marcin-michal (commented):
The error doesn't seem to be displayed in the step body, only in the header in the popup. If the rows are set to be autoexpended when an error occurs in them, it makes sense that the error would show up there.

I also don't think it's the right choice to use Tekton Results as a fallback if both k8s and Kubearchive queries fail. The user explicitly sets whether to use TR or Kubearchive to fetch logs in the feature flags, so to then quietly use TR seems wrong.

janaki29:
@marcin-michal Updated

Error in the step body — Failed steps still auto-expand, and the error is now shown inline under the section header (not only via the header tooltip icon).

Tekton Results fallback — Agreed. With kubearchive-logs enabled we shouldn’t silently switch to TR. That fallback is removed; if the pod can’t be loaded from k8s/KubeArchive we surface the fetch error instead.

PR #1191, opened 2026-07-17
fix(KFLUXUI-1528): fix TableV2 invalid flex column width

rrosatti:
## Fixes 

Fixes https://issues.redhat.com/browse/KFLUXUI-1528

## Description

When using flex type, the computed width could have a value not supported by PF's `Th` width prop (e.g. 17%), which was silently ignored - leaving columns with no width constraint. This caused column overlap and flickering during browser resize.

Fix: use `style={{ width: x% }}` directly instead of PF's `width` prop, so any percentage value works. This is consistent with how fixed-width columns already apply their widths.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Build related changes

Before:

https://github.com/user-attachments/assets/6685f5d2-f44f-4f66-867e-732b72e98d7c

AfteR:

https://github.com/user-attachments/assets/97c79e7d-864b-4b56-883e-1b3b7945ae98

## How to test or reproduce?

- Navigate to Snapshots tab (Applications -> select an app -> Snapshots)
- Resize the browser window horizontally
- Verify columns (Name, Created at, Components, etc.) don't overlap or flicker during resize

## Browser conformance: 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved table column sizing for flexible-width columns, ensuring the configured percentage is applied correctly.
  * Preserved existing fixed-width and sorting behavior.

* **Tests**
  * Updated coverage to verify flexible columns use the correct inline width styling.

### DIFF
--- a/src/shared/components/TableV2/TableHeader.tsx
+++ b/src/shared/components/TableV2/TableHeader.tsx
@@ -38,7 +38,7 @@ export const TableHeader = <TData,>({
             const widthProps: Partial<ThProps> = {};
 
             if (colWidth?.type === 'flex') {
-              widthProps.width = colWidth.widthPercent as ThProps['width'];
+              widthProps.style = { width: `${colWidth.widthPercent}%` };
             } else if (colWidth?.type === 'fixed') {
               widthProps.style = { width: colWidth.fixedWidth };
             }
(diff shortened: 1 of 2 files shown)


sahil143 (approved):

StanislavJochman (approved):

StanislavJochman (commented):
LGTM

PR #1299, opened 2026-08-05
feat(KFLUXUI-1399): instrument PipelineRuns list view with performance monitoring

sahil143:
## Fixes
Fixes: https://redhat.atlassian.net/browse/KFLUXUI-1399

## Description

Instruments the PipelineRuns list view (`PipelineRunsListViewV2`) with the performance monitoring hooks from the base PR in this stack.

Three hook calls added:
- **`useRenderTiming`** — tracks mount-to-interactive time (warn: 3s, critical: 8s)
- **`useLoadingThreshold`** for initial load — detects extended spinner (warn: 5s, critical: 15s)
- **`useLoadingThreshold`** for pagination — detects slow page fetches (warn: 3s, critical: 8s)

Spans appear in Sentry Performance waterfall, metrics in dashboards, threshold breaches as alertable events.

## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Build related changes

N/A — no UI changes. Adds monitoring instrumentation only.

## How to test or reproduce?

1. Enable Sentry on staging with `MONITORING_ENABLED=true` and `MONITORING_DSN`
2. Navigate to PipelineRuns list view
3. Check Sentry Performance for `pipelineruns.list.render` spans and metrics
4. Existing tests pass: `yarn test -- src/components/PipelineRun/PipelineRunListView/ --no-coverage` (18 tests)
5. Full suite: 484 suites / 4549 tests pass

## Browser conformance:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


sahil143:
### `Sentry.metrics.distribution` is being deprecated

`reportMetric` calls `Sentry.metrics.distribution()`. While it still works in v10.68, Sentry is moving away from the standalone metrics API — the official guidance is to use spans with attributes, which `startInactiveSpan` already does. So `reportMetric` is redundant with the span the hook already creates.

### `useRenderTiming` — has value, but simplify

Now that main has the full router integration (auto Web Vitals, page load/navigation transactions per route), this hook's use case is narrower but still valid: measuring "mount to interactive" for pages with multi-source data loading (e.g., PipelineRuns loading from cluster + KubeArchive/TektonResults simultaneously). Sentry's auto transaction uses idle timeout, not data completeness — this hook fills that gap.

**Suggestion:** Drop `reportMetric` from the provider interface. The span from `startInactiveSpan` already carries the timing data. Keep `useRenderTiming` but have it only use spans.

### `useLoadingThreshold` — reconsider

With the router integration now in place, Sentry Performance Alerts can be configured server-side to alert when specific route transactions exceed duration thresholds. That's more maintainable than setTimeout-based client-side alerting, and doesn't add per-page overhead. This hook is essentially reimplementing server-side alerting logic in the client.

### Missing: Global TanStack Query error handling

The biggest gap in the monitoring story isn't addressed by this PR or the base setup. `queryClient` in `src/k8s/query/core.ts` has no `QueryCache({ onError })`. Failed API calls that aren't explicitly caught by consuming components never reach Sentry. This would be higher impact than the custom hooks:

```ts
import { QueryClient, QueryCache } from '@tanstack/react-query';
import { monitoringService } from '~/monitoring';

export const queryClient = new QueryClient({
  queryCache: new QueryCache({
    onError: (error, query) => {
      monitoringService?.captureException(error, {
        queryKey: JSON.stringify(query.queryKey),
      });
    },
  }),
  defaultOptions: {
    queries: { staleTime: 30_000 },
  },
});
```

Zero per-page work, catches every failed query globally.

### Summary

| Item | Recommendation |
|------|---------------|
| `reportMetric` / `Sentry.metrics.distribution` | **Drop** — deprecated API, redundant with spans |
| `useRenderTiming` | **Keep** — simplify to use spans only, use for multi-source pages |
| `useLoadingThreshold` | **Reconsider** — server-side Sentry alerts are more maintainable |
| Global QueryCache error handler | **Add** — highest impact missing piece |

